PR #142, opened 2016-06-22
Correcting indentation for docs landing page.

drvinceknight:
Currently, on the [docs landing page](http://cosmic-ray.readthedocs.io/en/latest/) the enumerative list resets to `1` at each point. This is because the indentation isn't quite right in the markdown files.

Here is what this PR looks like locally:

screenshot from 2016-06-22 18-14-25

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -24,36 +24,35 @@ mutants, here's what you do:
 
 1. Install Cosmic Ray
 
-```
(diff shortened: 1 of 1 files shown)


abingham:
Excellent, thanks!

drvinceknight:
:) :+1:

PR #144, opened 2016-06-28
Allow session-name.json to be passed on the command line

atodorov:
This makes it easier to use bash tab completion


abingham:
Thanks! The "session" abstraction itself may go away at some point, but for now your change is a really useful, practical one.

PR #145, opened 2016-06-28
Report the mutation results using unified diff format

atodorov:
The idea comes from https://github.com/mbj/mutant - the Ruby mutation testing tool. I've tried to emulate their format as much as possible. I've also added the diff output in case there is an exception so it is easier to debug.  NOTE: The ast parser is stripping out comments, newlines and converting double to single quotes and because of this I'm diff'ing the unparsed value of the original and modifies AST trees. This also leads to difference in line numbers, but still should be easier to navigate than the N-th occurrence of operator X.

An example from one of my test executions:

```
job ID 81:Outcome.SURVIVED:pykickstart.commands.autopart
command: cosmic-ray worker pykickstart.commands.autopart replace_Eq_with_IsNot 3 unittest -- tests/
--- mutation diff ---
--- a/home/atodorov/repos/git/pykickstart/pykickstart/commands/autopart.py
+++ b/home/atodorov/repos/git/pykickstart/pykickstart/commands/autopart.py
@@ -268,7 +268,7 @@
         retval = F20_AutoPart.parse(self, args)
         if (self.fstype == 'btrfs'):
             raise KickstartParseError(formatErrorMsg(self.lineno, msg=_('autopart --fstype=btrfs is not valid fstype, use --type=btrfs instead')))
-        if ((self._typeAsStr() == 'btrfs') and self.fstype):
+        if ((self._typeAsStr() is not 'btrfs') and self.fstype):
             raise KickstartParseError(formatErrorMsg(self.lineno, msg=_('autopart --fstype cannot be used with --type=btrfs')))
         return retval
 RHEL7_AutoPart = F21_AutoPart

total jobs: 81
complete: 81 (100.00%)
survival rate: 100.00%
```

I'm not going to colorize the diff output at the moment. There is `colordiff` and other tools which appear to be working decently. In case you are desperate (or crazy) there is also a sed command listed at https://retracile.net/blog/2013/06/01/22.00.

### DIFF
--- a/cosmic_ray/commands/report.py
+++ b/cosmic_ray/commands/report.py
@@ -1,15 +1,23 @@
 def _print_item(item):
-    return [
-        'job ID: {}'.format(item.work_id),
-        'module: {}'.format(item.module_name),
-        'operator: {}'.format(item.operator_name),
-        'occurrence: {}'.format(item.occurrence),
+    result = item.result_data
+    result_type = item.result_type
+    if result_type in ['normal', 'exception']:
+        result_type = result[1][0]
+    ret_val = [
+        'job ID {}:{}:{}'.format(
+            item.work_id,
+            result_type,
+            item.module_name),
         'command: {}'.format(
             ' '.join(item.command)
             if item.command is not None else ''),
-        'result type: {}'.format(item.result_type),
-        'data: {}'.format(item.result_data)
         ]
+    if result_type == 'timeout':
+        ret_val.append("timeout: {:.3f} sec".format(result))
+    elif result_type in ['normal', 'exception']:
+        ret_val += result[1][1]
+
+    return ret_val
 
 
 def _get_kills(db):
--- a/cosmic_ray/testing/test_runner.py
+++ b/cosmic_ray/testing/test_runner.py
@@ -1,6 +1,8 @@
 import abc
 from collections import namedtuple
 from enum import Enum
+import sys
+import traceback
 
 
 class Outcome(Enum):
@@ -64,6 +66,6 @@ class TestRunner(metaclass=abc.ABCMeta):
             else:
                 return TestResult(Outcome.KILLED,
                                   test_result[1])
-        except Exception as e:
+        except Exception:
             return TestResult(Outcome.INCOMPETENT,
-                              str(e))
+                              traceback.format_exception(*sys.exc_info()))
--- a/cosmic_ray/worker.py
+++ b/cosmic_ray/worker.py
@@ -4,7 +4,10 @@ A worker is intended to run as a process that imports a module, mutates it in
 one location with one operator, runs the tests, reports the results, and dies.
 """
 
+import astunparse
+import difflib
 import importlib
+import inspect
 import logging
 import sys
 import traceback
@@ -12,6 +15,7 @@ import traceback
 from .importing import using_mutant
 from .mutating import MutatingCore
 from .parsing import get_ast
+from .testing.test_runner import TestResult, Outcome
 
 LOG = logging.getLogger()
 
@@ -58,21 +62,44 @@ def worker(module_name,
     """
     try:
         module = importlib.import_module(module_name)
+        module_source_file = inspect.getsourcefile(module)
         module_ast = get_ast(module)
+        module_source = astunparse.unparse(module_ast)
+
         core = MutatingCore(occurrence)
         operator = operator_class(core)
-        operator.visit(module_ast)
+        # note: after this step module_ast and modified_ast
+        # appear to be the same
+        modified_ast = operator.visit(module_ast)
+        modified_source = astunparse.unparse(modified_ast)
 
         if not core.activation_record:
             return ('no-test', None)
 
+        # generate a source diff to visualize how the mutation
+        # operator has changed the code
+        module_diff = ["--- mutation diff ---"]
+        for line in difflib.unified_diff(
+                        module_source.split('\n'),
+                        modified_source.split('\n'),
+                        fromfile="a"+module_source_file,
+                        tofile="b"+module_source_file,
+                        lineterm=""):
+            module_diff.append(line)
+
         with using_mutant(module_name, module_ast):
             results = test_runner()
+            # append the diff to whatever result was returned
+            res = results[1] or []
+            res.extend(module_diff)
+            results = TestResult(results[0], res)
 
         return ('normal',
                 (core.activation_record,
                  results))
 
     except Exception:
-        return ('exception',
-                traceback.format_exception(*sys.exc_info()))
+        res = traceback.format_exception(*sys.exc_info())
+        res.extend(module_diff)
+        results = [Outcome.INCOMPETENT, res]
+        return ('exception', (None, results))
--- a/cosmic_ray/cli.py
+++ b/cosmic_ray/cli.py
@@ -292,8 +292,6 @@ options:
             operator,
             int(config['<occurrence>']),
             test_runner)
-        if result_type == 'exception':
-            data = str(data)
 
     sys.stdout.write(
         json.dumps((result_type, data),
--- a/setup.py
+++ b/setup.py
@@ -71,6 +71,7 @@ setup(
     platforms='any',
     include_package_data=True,
     install_requires=[
+        'astunparse',
         'celery',
         'decorator',
         'docopt',


abingham:
I'm on vacation right now and can't look at this in great detail, but it looks pretty solid from what I can tell. A merge will have to wait until I get back home in about two weeks.

One note: I have vague plans to have a more sophisticated "reporting" infrastructure, and this would fall squarely in that domain. So we might have basic reports, summary reports, different reports, etc. It might even support plugin-based reporters. I guess what your patch highlights is the need for more information to be made available to the reporters.

atodorov:
@abingham - thanks for looking into this. I'd appreciate if you can merge it after you get back from holiday. A more sophisticated reporting framework, even plugin based, is something I fully support but don't really need at this moment. I'll be happy to refactor this patch later to fit into whatever you come up with.

atodorov:
Improved exception reporting

mbj:
Hi, Mutant author here. 

Are you people interested in a chat (hangout?), it might be interesting to share some other ideas on top of diff based reporting ;)

atodorov:
Hi @mjb,
I'm experimenting with mutation testing right now and have both Python and Ruby projects to work on. I'll be really interested in such a chat/whatever.

abingham:
I'm back from travel now, and I'll start looking at my pending patches soon. 

I'm definitely interested in some forum for discussing topics both for cosmic ray as well as the broader topics of mutation testing. My preference for these kinds of things is a text-based medium like a mailing list or slack, as opposed to talking (e.g. a hangout). Partly this is because I find the discussions easier to manage time-wise, but mostly this is because it means we have a searchable record of the discussion. What are your thoughts on this? Any preferences?

abingham:
@atodorov As I mentioned in #148, the defect mentioned there is actually a bug in this patch. The reporting system no longer accounts for the fact that `WorkItem.result_data` can have different types depending on the value of `WorkItem.result_type`. Specifically, in the case that a test times out, the type of `result_data` will be a float indicating how long the test ran. In other cases it's a more complex type.

We can definitely discuss whether the current design can be fixed, better documented, or whatever. But I'd prefer to take that topic up in a different issue. For this patch, I won't merge it in until the problem is fixed. Fortunately it shouldn't be too hard: `result_type` is "timeout" and treat `result_data` appropriately (i.e. as a float).

abingham:
I should add that, except for the handling of timeout results, this patch works great. I'm going to look closer at the code, but I don't foresee any problems merging this in as soon as the defect is addressed. Thanks!

cosmic_ray/testing/test_runner.py
```diff
@@ -1,4 +1,6 @@
 import abc
+import sys
+import traceback
```
abingham: I know it's a bit picky, but I'd prefer to keep import groups sorted alphabetically.

cosmic_ray/commands/report.py
```diff
@@ -1,15 +1,19 @@
...
-        'occurrence: {}'.format(item.occurrence),
+    result = item.result_data
+    result_type = item.result_type
+    if result_type in ['normal', 'exception']:
+        result_type = result[1][0]
+    ret_val = [
+        'job ID {}:{}:{}'.format(
+                                item.work_id,
+                                result_type,
+                                item.module_name),
```
abingham: Use PEP8 for this indentation, i.e. indent them 4 spaces in from the previous line. I'm not religious about this, but it lights my editor up when PEP8 is violated and I'll end up changing the lines myself at some point if you don't.

cosmic_ray/commands/report.py
```diff
@@ -1,15 +1,19 @@
...
         'command: {}'.format(
             ' '.join(item.command)
             if item.command is not None else ''),
-        'result type: {}'.format(item.result_type),
-        'data: {}'.format(item.result_data)
         ]
+    ret_val.extend(result[1][1])
```
abingham: It seems a bit strange to have this as an `extend()` rather than just adding it as part of the list literal above. Am I missing something?

cosmic_ray/testing/test_runner.py
```diff
@@ -66,4 +68,4 @@ def __call__(self):
                                   test_result[1])
         except Exception as e:
```
abingham: Since we're no longer using `e`, we don't need to capture it in the `except` clause.

cosmic_ray/worker.py
```diff
@@ -8,10 +8,14 @@
 import logging
 import sys
 import traceback
+import astunparse
+import difflib
+import inspect
```
abingham: Alphabetize these.

cosmic_ray/worker.py
```diff
@@ -58,21 +62,44 @@ def worker(module_name,
     """
     try:
         module = importlib.import_module(module_name)
+        module_source_file = inspect.getsourcefile(module)
```
abingham: This is brilliant! I did not realize that `inspect` had functions like `getsourcefile()` and `getsource()`. I think I can use these in other parts of CR.

cosmic_ray/worker.py
```diff
@@ -58,21 +62,44 @@ def worker(module_name,
...
         module_ast = get_ast(module)
+        module_source = astunparse.unparse(module_ast)
+
         core = MutatingCore(occurrence)
         operator = operator_class(core)
-        operator.visit(module_ast)
+        # note: after this step module_ast and modified_ast
+        # appear to be the same
```
abingham: Is this note still accurate?

mbj:
I've a slack channel, mutation-testing.slack.com, I invite anyone to who is interested. We right now discuss mutant, and general mutation testing topics.

I'm also happy to hand out administrator creds to other projects so this location becomes more populated. Ping me on intereested.

abingham:
@mbj I'm interested in the slack channel. Can you add me?

atodorov:
@mbj - add me on slack as well please.

@abingham - thanks for the review. I'll take a closer look in the upcoming days and update this PR.

cosmic_ray/testing/test_runner.py
atodorov: Maybe we can add pylint or something similar to analyze the code. If we do then removing the `as e` part will probably be flagged as warning. Although it will probably flag it as unused variable anyway. Removing the unused part for now.

cosmic_ray/worker.py
atodorov: I think yes. As far as I can tell the visit operation modifies the AST tree it operates on instead of returning another AST tree object. I've tried shuffling around some of the lines to prevent generating the diff if we don't have to and stumbled across this hence the comment.

cosmic_ray/testing/test_runner.py
atodorov: Side note: for some projects I'm using landscape.io as analytics tool. Here's the report for CR:
https://landscape.io/github/MrSenko/cosmic-ray/9/messages

I will go ahead and clean up some of these later.

cosmic_ray/worker.py
abingham: Ah, yes, you're absolutely correct. Sorry, I was a bit jet-lagged as I was going over these changes.

cosmic_ray/testing/test_runner.py
abingham: I'm currently running landscape on each commit to master, I think. At least, I've got a badge telling my I'm 83% healthy! I'll admit that I don't pay attention to those numbers very often, though.

mbj:
@abingham I need your email address, can you DM me on twitter? Or send an email to the address I have in my github account. https://twitter.com/_m_b_j_

@atodorov Invited to the email listed in your GH profile.

Lets create a #cosmic-ray channel once you are in. And I'm happy to grant admin rights so more people can be invited independently to me. We did the same for the JS mutation testing tool https://github.com/stryker-mutator/stryker, that is not as active than the mutant community right now.

cosmic_ray/commands/report.py
```diff
@@ -1,15 +1,23 @@
...
             if item.command is not None else ''),
-        'result type: {}'.format(item.result_type),
-        'data: {}'.format(item.result_data)
         ]
+    if result_type == 'timeout':
+        ret_val['timeout'] = '{:.3f} sec'.format(result)
+    else:
+        ret_val += result[1][1]
+
+    return ret_val
```
atodorov: This is my fix for the timeout value. Should be good to go.

I'd love to see some unified results format & more docs but that's definitely another PR.

cosmic_ray/commands/report.py
```diff
@@ -1,15 +1,23 @@
...
+            result_type,
+            item.module_name),
         'command: {}'.format(
             ' '.join(item.command)
             if item.command is not None else ''),
-        'result type: {}'.format(item.result_type),
-        'data: {}'.format(item.result_data)
         ]
+    if result_type == 'timeout':
+        ret_val['timeout'] = '{:.3f} sec'.format(result)
```
abingham: What's going on here? It looks like `ret_val` is a list, but you're treating it like a dict. Perhaps you meant something like `ret_val.append("timeout: {:.3f} sec".format(result))`.

cosmic_ray/commands/report.py
```diff
@@ -1,15 +1,23 @@
...
+    else:
+        ret_val += result[1][1]
```
abingham: I'm a bit worried about this line. I can't run CR where I currently am, so I can't verify my concerns. Will `result` always be indexable in this way? I think it will be OK if `result_type` is "normal" or "exception", but not if it's "no-test" (which shouldn't generally happen). 

Also, since you're using `+=`, you're assuming that `result[1][1]` is an iterable object. Will that always be the case? Or should you instead be using `ret_val.append()` here?

I'll try to look at this more closely this evening when I get home.

cosmic_ray/commands/report.py
atodorov: Argh, thanks for catching this.

cosmic_ray/commands/report.py
atodorov: Should be better with the latest update. 

Wrt `+=` and `.append()` - if I use append then I get a list inside a list.

PR #146, opened 2016-07-07
Accept directories and .py files as module names

atodorov:
This is another minor improvement patch which lets you do things like:

    cosmic-ray init --baseline=10 authconfig pykickstart/commands/authconfig.py -- tests/

I find it very handy b/c I'm using tab-completion all the time.

### DIFF
--- a/cosmic_ray/modules.py
+++ b/cosmic_ray/modules.py
@@ -3,11 +3,26 @@
 
 import importlib
 import logging
+import os
 import pkgutil
 import re
 
 LOG = logging.getLogger()
 
+def find_local_modules(name, excludes=None):
+    """
+        Just calls find_modules() but taking into
+        account the fact that name may be a relative or full
+        path to a .py file or directory.
+    """
+    # helps when using bash completion
+    if os.path.exists(name):
+        if name.endswith('.py'):
+            name = name[:-3]
+        name = name.replace('/', '.')
+
+    for module in find_modules(name, excludes):
+        yield module
 
 def find_modules(name, excludes=None):
     """Generate sequence of all submodules of NAME, including NAME itself.
--- a/cosmic_ray/test/test_find_modules.py
+++ b/cosmic_ray/test/test_find_modules.py
@@ -1,6 +1,6 @@
-import contextlib
-import os.path
 from pathlib import Path
+import contextlib
+import os
 import sys
 
 import cosmic_ray.modules
@@ -15,16 +15,20 @@ def extend_path(directory):
     """Put `directory` at the front of `sys.path` temporarily.
     """
     sys.path = [str(directory)] + sys.path
+    old_dir = os.getcwd()
+    os.chdir(str(directory))
     try:
         yield
     finally:
         sys.path = sys.path[1:]
+        os.chdir(old_dir)
 
 
 def test_small_directory_tree():
     datadir = THIS_DIR / 'data'
     paths = (('a', '__init__.py'),
              ('a', 'b.py'),
+             ('a', 'py.py'),
              ('a', 'c', '__init__.py'),
              ('a', 'c', 'd.py'))
     expected = sorted(datadir / Path(*path) for path in paths)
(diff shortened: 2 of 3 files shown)


abingham:
My only concern here is that I could have a file named e.g. "foo/py.py". In this case "foo.py" would be somewhat ambiguous...clearly that filename doesn't exist, but this code would parse it as "foo". 

Perhaps we could let the user specify an optional flag to force filename or module name interpretation.

Otherwise I really like this change. It's precisely the kind of polish the project needs.

atodorov:
I've reverted my changes and added a test for this edge case
```
def test_finding_modules_foo_dot_py_using_dots():
     datadir = THIS_DIR / 'data'
    paths = (('foo', 'py.py'),)
     expected = sorted(datadir / Path(*path) for path in paths)
     with extend_path(datadir):
         results = sorted(
             Path(m.__file__)
            for m in cosmic_ray.modules.find_modules('foo.py'))
     assert expected == results
```

it fails with 
```
______________________________________________ test_finding_modules_foo_dot_py_using_dots _______________________________________________

    def test_finding_modules_foo_dot_py_using_dots():
        datadir = THIS_DIR / 'data'
        paths = (('foo', 'py.py'),)
        expected = sorted(datadir / Path(*path) for path in paths)
        with extend_path(datadir):
            results = sorted(
                Path(m.__file__)
                for m in cosmic_ray.modules.find_modules('foo.py'))
>       assert expected == results
E       assert [PosixPath('/...a/foo/py.py')] == []
E         Left contains more items, first extra item: PosixPath('/home/senko/cosmic-ray/cosmic_ray/test/data/foo/py.py')
E         Use -v to get the full diff

cosmic_ray/test/test_find_modules.py:66: AssertionError
--------------------------------------------------------- Captured stderr call ----------------------------------------------------------
Unable to import foo.py
Traceback (most recent call last):
  File "/home/senko/cosmic-ray/cosmic_ray/modules.py", line 52, in find_modules
    module = importlib.import_module(module_name)
  File "/home/senko/.virtualenvs/CosmicRay/lib64/python3.5/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 986, in _gcd_import
  File "<frozen importlib._bootstrap>", line 969, in _find_and_load
  File "<frozen importlib._bootstrap>", line 944, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 222, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 986, in _gcd_import
  File "<frozen importlib._bootstrap>", line 969, in _find_and_load
  File "<frozen importlib._bootstrap>", line 956, in _find_and_load_unlocked
ImportError: No module named 'foo'
```

So we don't really handle this case very well or my test code is bogus. Please advice. I can make both of them work in this PR.

abingham:
I'm back from vacation in a few days. I'll take a look at this then.

abingham:
I think you need to create a "foo/__init__.py" as well. Otherwise foo won't be a package.

atodorov:
Please check my latest updates. I've added a py.py file in the test directory and also added a few more tests. I'm doing the .py suffix and / replacements only if the given name is a directory or a file on disk. In the tests I had to mock the os.path.exists method, I guess b/c of the relative path where the data files are stored.

cosmic_ray/modules.py
```diff
@@ -5,6 +5,7 @@
 import logging
 import pkgutil
 import re
+import os
```
abingham: alphabetize the imports

cosmic_ray/modules.py
```diff
@@ -35,6 +36,11 @@ def find_modules(name, excludes=None):
 
     """
     excludes = excludes or []
+    # helps when using bash completion
+    if os.path.exists(name):
+        if name.endswith('.py'):
+            name = name[:-3]
+        name = name.replace('/', '.')
```
abingham: I would really prefer this logic to be in cli.py. This feels more like a "creature comfort" for command line users than it does a part of the `find_modules()` algorithm.

cosmic_ray/test/test_find_modules.py
```diff
@@ -1,14 +1,18 @@
-import contextlib
-import os.path
 from pathlib import Path
+import contextlib
+import pytest
 import sys
+import os
```
abingham: alphabetize these.

abingham:
It really bothers me that we don't know my `os.path.exists()` doesn't work as expected in the current patch. If we move the special ".py" handling into cli.py, then the problem goes away, though.

cosmic_ray/modules.py
atodorov: Can you point me to the right place in the code? I see that cli.py calls find_modules twice and I don't want to repeat the same code twice.

atodorov:
> It really bothers me that we don't know my os.path.exists() doesn't work as expected in the current patch. 

I think this is because the test modules are under `THIS_DIR / data` which is temporarily added the sys.path, while `os.path.exists(<relative_path>)` looks for files and directories under the current working directory, not under `THIS_DIR / data`

abingham:
> os.path.exists(<relative_path>) looks for files and directories under the current working directory, not under THIS_DIR / data

Ah, right, I see what you're saying now. In that case, I think it would be cleaner to simply change to `THIS_DIR/data` before running `find_modules()`. This way we're actually testing all of the functionality, not faking some of it.

We used to have a context manager for doing temporary directory switches:

```
@contextlib.contextmanager
def excursion(directory):
    old_dir = os.getcwd()
    os.chdir(directory)
    try:
        yield
    finally:
        os.chdir(old_dir)
```

Just put this into `test_find_modules.py` for now.

cosmic_ray/modules.py
abingham: I'd create a helper function in `cli.py` which contains the code. Just something like:
```
def _fixup_module_name(name):
    if os.path.exists(name):
        if name.endswith('.py'):
            name = name[:-3]
        name = name.replace('/', '.')
    return name
```

Then, in the two places where we use `find_modules()`, use this function to adjust the module name, e.g. `_fixup_module_name(configuration['<top-module>'])`.

Alternatively, create a complete wrapper for `find_modules()` which does this fixup, and then call this wrapper in place of the `find_modules()`. Whichever approach seems cleaner to you.

atodorov:
This one should be good to go once the tests pass

PR #153, opened 2016-07-13
Return a list of strings for unittest results, not list of tuples

atodorov:
This is needed so the reporter can print a nicely formatted traceback when the job is killed.

You can use the Python examples at https://github.com/atodorov/mutation-testing-example to see how results are reported with and w/o this patch. 

Also I'm not sure how the pytest runner will report the failed results and if that wouldn't clash with this existing patch. I don't have a pytest suite ATM but I'll try to make one a give it a try.

### DIFF
--- a/cosmic_ray/testing/unittest_runner.py
+++ b/cosmic_ray/testing/unittest_runner.py
@@ -22,6 +22,4 @@ class UnittestRunner(TestRunner):  # pylint:disable=no-init, too-few-public-meth
 
         return (
             result.wasSuccessful(),
-            [(str(r[0]), r[1])
-             for r in chain(result.errors,
-                            result.failures)])
+            [r[1] for r in chain(result.errors, result.failures)])


abingham:
Why is this change necessary? As far as I can tell you haven't done anything but remove data from the test runner results. All of the data you need for the traceback formatting was already there.

atodorov:
Right, I'm removing data b/c in `worker.py` lines 93-94 we assume that the result type is a list of strings, not a list of tuples. Without this patch we'd have to check the type of `res` and special case the results coming from `unittest_runner`.

abingham:
I don't see any assumption like that in the code you mentioned; lists are heterogeneous, etc.. Do you mean that the report generator makes assumption about the data being a list of strings? That may be the case, but if so it's an error.

I think I see where you're going with this, and really it points to the larger (and very valid) question of how TestRunners should structure their data. Right now, I leave that structure almost entirely up to the TestRunner implementation by saying that it should return

> any object that is appropriate to provide more information about the success/failure of the tests.

(cosmic_ray/testing/test_runner.py:44)

As such, any report generator will not be able to make any assumptions about how the test result data is shaped. If I was to accept this patch, it would imply (to my mind, at least) that we'd defined a canonical test result structure and, as such, we'd need to document it and update any existing test runners. I don't think we're there yet, so I'm not really inclined to accept this patch yet.

Right now I see three ways to approach this issue. First, we could ignore it and simply rely on test runners returning something with a reasonable repr() and/or str() format. This is more or less what we do now, and it has certain benefits.

Second, we could define a strict structure for test result data. This would probably work in practice; I imagine we could find something sufficiently universal for all potential test runners that also contains enough information to be useful.

Third, we could support some notion of "test result formatter" plugins that know how to handle the data for the various test runners. This feels a bit overengineered right now, but might be the correct solution in the long term.

I suppose a fourth option, related to the third, is to support report plugins as we've discussed elsewhere. You could then have different reporters for different test result structures. This would likely result in significant duplication of code and/or divergent report formats, though, and I don't really like it.

abingham:
FWIW, I think the second option - a prescribed test result format - is the pragmatic way to go for now. And since we know that the results have to be serialized through JSON to get to stdout, a list of strings is probably about as general as we can get...so perhaps your patch is going in the right direction! 

What are your thoughts on this?

atodorov:
I think I also like the second option the most for the time being. Test result formatter sounds also nice but let's make that happen when it's really needed. ATM I'm not sure if I really need anything else than the test output and a pass/fail result.

abingham:
I'll go ahead and merge this in for now. We can think about the "proper" solution as we get more experience using the tool for real. Thanks for your work on this!

PR #154, opened 2016-07-13
Don't show test output and mutation diff if result is KILLED

atodorov:
this makes it possible to keep cleaner output and only reports
diffs for mutations which survived or were incompetent. Use the
--full-report option to show everything if you want.

We can probably extend this to other types of results as well. For now I only want to see the diff for mutants that survived so I can easily spot where I'm missing test cases.

Another possibility is to report only the survived mutants and nothing else. What do you think ?

### DIFF
--- a/cosmic_ray/cli.py
+++ b/cosmic_ray/cli.py
@@ -192,16 +192,20 @@ options:
 
 
 def handle_report(configuration):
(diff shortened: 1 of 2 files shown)


abingham:
This looks fine. I figure we'll rework the whole presentation layer at some point, so it's good to experiment like this for now to see what we like, etc.

PR #155, opened 2016-07-13
Add Boolean replacer

atodorov:
@abingham I've fixed a minor issue with the `--full-report` option which you probably want to cherry-pick.

Next I've enabled reports in CI to see what's going on wrong and added a Boolean replacement operator which simply flips `True` and `False` values.

ATM I'm missing tests in `cosmic_ray/test/test_operators.py` b/c I'm not quite sure what is the format.
Also the adam tests are failing b/c there is a `while True:` loop which becomes `while False`. I can modify the loop somehow so we always trigger an infinite loop and not conflict with the BooleanReplacer operator or maybe you can tell me what other tests to add.

### DIFF
--- a/test_project/adam.py
+++ b/test_project/adam.py
@@ -7,6 +7,13 @@
 def constant_number():
     return 42
 
+def constant_true():
+    return True
+
+
+def constant_false():
+    return False
+
 
 def unary_sub():
     return -1
--- a/test_project/adam.py
+++ b/test_project/adam.py
@@ -47,5 +47,9 @@ def use_continue(limit):
 def trigger_infinite_loop():
     # When `break` becomes `continue`, this should enter an infinite loop. This
     # helps us test timeouts.
-    while True:
+    # Any object which isn't None passes the truth value testing so here
+    # we use `while object()` instead of `while True` b/c the later becomes
+    # `while False` when BooleanReplacer is applied and we don't trigger an
+    # infinite loop.
+    while object():
         break
--- a/cosmic_ray/operators/boolean_replacer.py
+++ b/cosmic_ray/operators/boolean_replacer.py
@@ -1,0 +1,24 @@
+import ast
+
+from .operator import Operator
+
+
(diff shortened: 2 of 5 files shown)


test_project/adam.py
```diff
@@ -7,6 +7,9 @@
 def constant_number():
     return 42
 
+def constant_boolean():
```
abingham: I think we should have `constant_true()` and `constant_false()` functions (with associated tests) so that we can verify behavior in both directions.

abingham:
Regarding `test_operators.py`, what you need to do is add an entry to `OPERATOR_SAMPLES` containing a) the operator to be tested and b) some code that it can modify. So something like `(BooleanReplacer, "True")`. Basically, the tests loop over that list and ensure that each operator seems to do something sensible. 

They're not perfect tests by any means, but they make sure the plumbing seems sane.

atodorov:
@abingham - I think I fixed it for now. `while True` becomes `while object()`, see the comments in the code.

test_project/run_tests.sh
```diff
@@ -4,13 +4,13 @@
 set -e
 
 cosmic-ray load cosmic-ray.unittest.conf
-#cosmic-ray report adam_tests.unittest
+cosmic-ray report adam_tests.unittest
```
abingham: Just re-comment these calls to report, and I'll be happy to merge this in.

test_project/adam.py
```diff
@@ -40,5 +47,9 @@ def use_continue(limit):
 def trigger_infinite_loop():
     # When `break` becomes `continue`, this should enter an infinite loop. This
     # helps us test timeouts.
-    while True:
+    # Any object which isn't None passes the truth value testing so here
+    # we use `while object()` instead of `while True` b/c the later becomes
+    # `while False` when BooleanReplacer is applied and we don't trigger an
+    # infinite loop.
+    while object():
```
abingham: This is a good idea!

Issue #157, opened 2016-07-14
CR fails to kill a mutant when the mutated module is loaded via importlib


atodorov:
Please see this for reproducer:
https://github.com/atodorov/mutation-testing-example/commit/8f8a3ab4247b9fcb9bb4f5abfbd4da1300d13571

I have a `sandwich/ham/ham.py` module. The `Ham` class is loaded in `sandwich/control.py` using `importlib.import_module`. Note that the loaded module is in a subdirectory and we modify `sys.path` - this seems to be related to the problem.

Then I have `test_control.py` which grabs the `ham_class` from `control.py` and creates an object of this type, then proceeds to assert on the default values of this object.

The execute the test suite manually run `python3 -m unittest tests/test_control.py`  - everything is fine.
If I make the below mutation by manually editing the code and re-run the test suite it fails as expected:

```
$ python3 -m unittest tests/test_control.py 
F
======================================================================
FAIL: test_loading_via_importlib (tests.test_control.TestControl)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/atodorov/mutation-testing-example/tests/test_control.py", line 7, in test_loading_via_importlib
    self.assertEqual(ham_in_fridge.pieces, 10)
AssertionError: 11 != 10

----------------------------------------------------------------------
Ran 1 test in 0.000s

FAILED (failures=1)
```

However when running CR the very same mutation leads to a surviving mutant:

```
$ cosmic-ray run --baseline=10 ham.json sandwich/ham/ham.py -- tests

$ cosmic-ray report ham.json
job ID 1:Outcome.SURVIVED:sandwich.ham.ham
command: cosmic-ray worker sandwich.ham.ham number_replacer 0 unittest -- tests
--- mutation diff ---
--- a/home/atodorov/mutation-testing-example/sandwich/ham/ham.py
+++ b/home/atodorov/mutation-testing-example/sandwich/ham/ham.py
@@ -3,6 +3,6 @@

 class Ham(object):

-    def __init__(self, pieces=10):
+    def __init__(self, pieces=11):
         self.pieces = pieces

total jobs: 1
complete: 1 (100.00%)
survival rate: 100.00%
```

abingham:
Ah, this is interesting. This could be complex-to-impossible, and maybe it'll require monkey patching to be completely watertight. I'm not sure when I'll have time to really look into it, but hopefully soon.

abingham:
[...]

PR #158, opened 2016-07-20
Delete parent modules from sys.modules

atodorov:
I believe this fixes my comment at
https://github.com/sixty-north/cosmic-ray/issues/157#issuecomment-233741363

Here is the test log w/o the fix and then with the fix:

https://travis-ci.org/MrSenko/cosmic-ray/jobs/146035928

https://travis-ci.org/MrSenko/cosmic-ray/jobs/146038003

### DIFF
--- a/cosmic_ray/importing.py
+++ b/cosmic_ray/importing.py
@@ -4,6 +4,7 @@
 import contextlib
 from importlib.abc import MetaPathFinder
 from importlib.machinery import ModuleSpec
+from itertools import accumulate
 import sys
 
 
@@ -61,8 +62,11 @@ def using_mutant(module_name, mutant):
     mutated AST.
 
     """
-    if module_name in sys.modules:
-        del sys.modules[module_name]
+    # delete the module we're mutating and all of it's parent modules
+    # this will force a re-import of the module chain
+    del_mods = set(accumulate(module_name.split('.'), '{}.{}'.format))
+    for k in del_mods:
+        del sys.modules[k]
 
     finder = ASTFinder(module_name, mutant)
     sys.meta_path = [finder] + sys.meta_path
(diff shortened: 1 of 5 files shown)


atodorov:
FYI this also fixes my problem in #157 where I modify `sys.path` and load modules via `importlib.import_module`, at least in the production SUT, but not in the example. I guess in the larger piece of software removing the parent modules from `sys.modules` is causing them to reload the children and I get the mutated ast loaded. However the same thing doesn't happen with the initial example I've provided.

cosmic_ray/importing.py
```diff
@@ -61,8 +61,11 @@ def using_mutant(module_name, mutant):
     mutated AST.
 
     """
-    if module_name in sys.modules:
-        del sys.modules[module_name]
+    module_to_delete = ""
+    for parent in module_name.split('.'):
+        module_to_delete = ("%s.%s" % (module_to_delete, parent)).lstrip('.')
+        if module_to_delete in sys.modules:
+            del sys.modules[module_to_delete]
```
abingham: This is a bit awkward and old-fashioned. I'd prefer something a bit more modern and (to me, at least) readable:
```
del_mods = set(accumulate(module_name.split('.'), '{}.{}'.format))
sys.modules = {k: v for k, v in sys.modules.items() if k not in del_mods}
```

atodorov: For some reason the assignment to sys.modules doesn't have the desired effect. Looping through del_mods and deleting them one by one however seems to be fine.

abingham: Interesting. Something must be holding a reference to the original.

abingham:
I'll admit that I don't entirely understand the underlying issue here - in particular, why `from sandwich.ham import ham` doesn't invoke the finder when `sandwich.ham.ham` isn't in `sys.modules` - but this patch does seem to address at least part of the issue.

PR #160, opened 2016-07-20
Workaround for empty __init__.py files

atodorov:
Running against one of my production SUT I was seeing an error similar to this:
https://travis-ci.org/MrSenko/cosmic-ray/jobs/146085309

With the workaround everything seems to work fine:
https://travis-ci.org/MrSenko/cosmic-ray/jobs/146087865

Note the upstream Python bug http://bugs.python.org/issue27578. I doubt it will get fixed or at least not soon enough, hence the need to workaround.

### DIFF
--- a/cosmic_ray/parsing.py
+++ b/cosmic_ray/parsing.py
@@ -22,15 +22,20 @@ def get_ast(module):
         TypeError: If the source file for `module` can't be found.
     """
     try:
-        source = inspect.getsource(module)
-    except OSError:
-        LOG.warning("Unable to read source for module %s", module)
-        raise
-
-    try:
         source_file = inspect.getsourcefile(module)
     except TypeError:
         LOG.error("Unable to get source file for object %s", module)
         raise
+
+    try:
+        source = inspect.getsource(module)
+    except OSError:
+        # workaround for empty __init__.py files, see
+        # http://bugs.python.org/issue27578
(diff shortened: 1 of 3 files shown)


cosmic_ray/parsing.py
```diff
@@ -22,15 +23,21 @@ def get_ast(module):
...
     try:
-        source = inspect.getsource(module)
-    except OSError:
-        LOG.warning("Unable to read source for module %s", module)
-        raise
-
-    try:
         source_file = inspect.getsourcefile(module)
     except TypeError:
+        source_file = ''
```
abingham: This seems unnecessary since we're not going to do anything with `source_file` afterward, i.e. we're just going to re-raise the exception and exit the function.

cosmic_ray/parsing.py
```diff
@@ -22,15 +23,21 @@ def get_ast(module):
...
         LOG.error("Unable to get source file for object %s", module)
         raise
 
+    try:
+        source = inspect.getsource(module)
+    except OSError:
+        # workaround for empty __init__.py files, see
+        # http://bugs.python.org/issue27578
+        if source_file.endswith('__init__.py') and os.path.exists(source_file):
+            source = open(source_file, 'r').read()
```
abingham: Won't `source` always be an empty string in this case? If so, we don't need to open and read the file.

PR #162, opened 2016-07-25
Skip  Eq and Is  replacement

atodorov:
In my experience these two appear to be equivalent and it makes sense to skip this replacement in the same fashion we do for their Not counterparts.

@abingham what do you think ?

### DIFF
--- a/cosmic_ray/operators/relational_operator_replacement.py
+++ b/cosmic_ray/operators/relational_operator_replacement.py
@@ -23,6 +23,7 @@ RELATIONAL_OPERATORS = {ast.Eq, ast.NotEq, ast.Lt, ast.LtE, ast.Gt, ast.GtE,
 # they almost always produce equivalent mutants. This is a set of
 # (FROM-OP, TO-OP) tuples which we don't want to generate.
 SKIP = {
+    (ast.Eq, ast.Is),
     (ast.NotEq, ast.IsNot),
 }
 


abingham:
(Sorry, this turned into a bit of a "think out loud" session)

I had thought about this one earlier, but I wasn't convinced that `==` and `is` are equivalent enough. Fundamentally, they're very different things, and - unlike the existing `!=`/`is not` exception - `is` is not an almost perfect superset of `==`.

(I can think of only one common case where `is not` is not the same as `!=`:
```
>>> x = float('NaN')
>>> x != x
True
>>> x is not x
False
```
While you technically _can_ create classes that exhibit this behavior, it's such an oddity that many programmers have to be continually reminded that NaNs are never equal. Which is to say that I think `!=` and `is not` are equivalent enough for our purposes right now.)

On the other hand, I can think of a vast number of examples where `is` would not have the same result as `==`. Any class that defines `__eq__` in a conventional manner would likely break this equivalence. This probably _is_ and argument for adding an exception for `is`/`==` replacement...things that are the same are generally equal.

Where things get really painful is with e.g. numbers. Consider:
```
>>> x = 1
>>> y = 1
>>> y == x
True
>>> y is x
True
>>> a = 12345678
>>> b = 12345678
>>> a == b
True
>>> a is b
False
>>>
```
These kinds of relationships are implementation-specific, probably version-specific, not promised or documented, and possibly dynamic. So no matter what strategy we choose WRT exceptions, we're in a hole.

Ultimately, I think the skip set is an interim solution. We'll _really_ solve this with a robust exceptions system that lets users indicate when to avoid certain operators. We might also broaden the operator API e.g. by letting operator developers add metadata which could be filtered on. For instance, certain operators could be marked "often equivalent" or something, and users could choose to not apply those at all.

abingham:
All of which is to say, I'll take this patch since it's probably really convenient for you and other users right now. I'm going to add a note to the skip set, though, indicating that it's an imperfect interim solution. Thanks for all of your work on this. It's really is great to get collaborators!

proggga:
I think we should add some exclusions
I have example, which is really valid for test, 
because logic doesn't changed, but I don't know, how it can be fixed:
i have comprehension like:

`[i for i in list if i is not None]`

cosmic rays change it to 

`[i for i in list if i != None]`

and my test doesn't fail, how I should pass this situation

abingham:
@proggga I agree that this mutation would be difficult for tests for detect, so another temporary exception is in order. Can you put together a PR for that?

If possible, I'd like the exception to be as tight as possible, only preventing mutation when `None` is one of the arguments to `is not`. I think this is possible, but I might be wrong.

proggga:
I'm sorry, but what you mean under "put together a PR", I will be happy to help with this project

abingham:
That just means to put create a *P*ull *R*equest with the appropriate changes. We'll look them over, maybe requests changes, and then merge it in. You can read more about PRs in the [github documentation](https://help.github.com/articles/about-pull-requests/).

(If you already know all of this, I don't mean to be rude or anything...I'm just not sure how much you know yet. I'll be happy to help you through the process if you need.)

PR #163, opened 2016-07-28
Added additional python versions.

degustaf:
Python 3.5, 3.5-dev, and nightly build were added, and the tests pass.
Python 3.3 required a change of version for celery, and is still failing.  But could be salvageable.
Python 3.2 does not support `yield from`, so it is unlikely that it is worth the effort to make this work.


abingham:
That looks really good, thanks!

Issue #101, opened 2015-04-18
Support more versions of python


abingham:
Right now we only support Python 3 (and probably really only Python 3.3+ or something like that.) It may not be easy to support multiple versions of Python in a single code base; for example, the finder/loader system may not be available very far back. In this case perhaps we should just have two version of the tool.
---
- Bitbucket: https://bitbucket.org/austin_bingham/cosmic-ray/issue/3
- Originally reported by: [austin_bingham](http://bitbucket.org/austin_bingham)
- Originally created at: 2015-02-04T12:59:08.661

DRMacIver:
FWIW, I think being Python 3 only is entirely reasonable for source transformation tools. Python 3.2 support would be nice because that's the latest version pypy supports (even if it *is* a horrible limited version of the language)

abingham:
Yeah, I'm not really motivated to support Python 2. I think I'd only consider that if I got a lot of feedback from users that wanted it. 3.2 is probably a good idea, though.

jonafato:
Feel free to disregard the following, as it is just an opinion. 3.2 support is likely not worth it. Per [this recent report](https://caremad.io/2015/04/a-year-of-pypi-downloads/):

> For authors wondering what versions of Python 3.x they should support, it’s easy to make the argument that targeting 3.4, or at the very least 3.3 is a completely reasonable target given that 3.2 (and to a lesser extent, 3.3) have a very tiny marketshare which very likely mostly consists of automated CI systems.

The linked stats show a < 0.5% install rate for Python 3.2 packages, and a substantial majority is coming from 3.4. This is not to say that users of 3.2 are not important, but support probably shouldn't take precedence over development of the project on other fronts, especially as users upgrade to current versions.

abingham:
Probably a reasonable first step here would be to figure out *which* versions of Python3 already work for CR :)

degustaf:
I just made a pull request (#163) that should help with this.  travis is now running tests for Python 3.3 and newer, and the Python 3.3 tests are allowed to fail.

atodorov:
@abingham based on the comments in #163 and #165 I think we can safely close this issue. Python 3.2 and 3.3 support doesn't appear to be possible (at least not easily) and maybe not even desirable at this moment.

abingham:
I agree.

snarfed:
[...]

PR #165, opened 2016-09-08
Python 3.3 fixes and AND/OR mutation operator

atodorov:
I am submitting these as one PR because I've set out to make the BoolConstReplacer operator 3.3 compatible and then run into some issues.

From #163 

> Python 3.3 required a change of version for celery, and is still failing. But could be salvageable.

@abingham, @degustaf - in `cosmic_ray/importing.py` we make use of `importlib.machinery.ModuleSpec` which was introduced in 3.4. I don't think we can/want to backport this to 3.3 so maybe drop the entire 3.3 compatibility mode ? What do you say ?

@abingham - do you have any preference if I should drop my 3.3 related commits since they will be kind of pointless if we drop this version entirely ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,13 +2,10 @@ language: python
 
 matrix:
   include:
-    - python: 3.3
     - python: 3.4
     - python: 3.5
     - python: 3.5-dev
     - python: nightly
-  allow_failures:
-  - python: 3.3
   fast_finish: true
 
 install: python setup.py install
--- a/cosmic_ray/cli.py
+++ b/cosmic_ray/cli.py
@@ -3,7 +3,6 @@
 Here we manage command-line parsing and launching of the internal
 machinery that does mutation testing.
 """
-from contextlib import redirect_stdout
 import itertools
 import json
 import logging
@@ -24,6 +23,7 @@ import cosmic_ray.json_util
 import cosmic_ray.worker
 import cosmic_ray.testing
 import cosmic_ray.timing
+from cosmic_ray.util import redirect_stdout
 from cosmic_ray.work_db import use_db, WorkDB
 
 
--- a/cosmic_ray/operators/boolean_replacer.py
+++ b/cosmic_ray/operators/boolean_replacer.py
@@ -54,3 +54,52 @@ class ReplaceOrWithAnd(Operator):
     def mutate(self, node):
         """Replace OR with AND."""
         return ast.And()
+
+
+class RemoveNot(Operator):
+
+    """An operator that removes the 'not' keyword from expressions."""
+
+    def visit_UnaryOp(self, node):  # noqa
+        """
+        http://greentreesnakes.readthedocs.io/en/latest/nodes.html#UnaryOp
+        """
+        if isinstance(node.op, ast.Not):
+            return self.visit_mutation_site(node)
+        else:
+            return node
+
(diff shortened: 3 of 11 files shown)


degustaf:
I have no problem leaving this incompatible with 3.3. In fact, that sounds like the right idea.

I just didn't (don't) understand the internals well enough to make that determination.

abingham:
TL;DR - I'm happy to drop 3.3- support.

I figure that CR is far enough from "complete" that I don't want to spend a lot of time making it compatible with Python versions that will be effectively dead when CR is finally fully armed and operational. We have to muck around with fairly obscure elements of the language that seem likely to change, and we could expend a lot our limited time and energy on mixed compatibility.

Based on that and [this discussion](https://github.com/sixty-north/cosmic-ray/issues/101), I'm happy to limit our backwards compatibility arbitrarily. It should work on 3.5, and 3.4 would be very nice. 3.3 and earlier feels unnecessary (@DRMacIver's points about PyPy being a reasonable counterargument).

DRMacIver:
I actually don't think it's a reasonable counterargument any more! The version of pypy that supports 3.2 was super old and didn't really work properly, and there's a new and fairly well funded effort which should upgrade the version of Python supported. Right now the next version of it will be supporting 3.3, but honestly I'd just wait and see where they get to before worrying about pypy (if you're not planning to support Python 2, which is totally reasonable).

atodorov:
I have removed Python 3.3 from the test matrix but otherwise left the other commits in case we want to revisit 3.3 support sometime in the future.

atodorov:
@abingham sorry for making this PR too big but some of the changes go together.

I have added two more mutation operators. One that removes the not keyword and another one which adds it in front of any test condition (if, assert, while and ifexp). Few notes:

* Travis CI fails due to stevedore not being available for Python nightly.
* For some reason the py.test runner fails with 3 survivors while the previous runner did not. All 3 survivors raise an exception and I suspect there is some problem with that. Will look into it so you can merge this without breaking the master branch.

* in `test_mutation_changes_ast` I have commented out the length comparison of original and mutated nodes b/c the not add/remove operators will change that. I'm not sure if we want to keep track of how many new nodes were added or removed. Maybe file an issue if we want to do it.
* The not keyword is added in front of any test conditions. However it is not added before operands of BoolOp, e.g. A or B will not become A or not B. ATM I don't think this is possible to do. I'd love to have some way to mutate first the left side, then the right side of the operation. Maybe that's another RFE/issue.


Please comment on the latest updates and let me know how to proceed.

degustaf:
@atodorov Looking at the [travis-ci build log](https://travis-ci.org/sixty-north/cosmic-ray/jobs/158751677#L353), it apears that the issue is that stevedore isn't compatible with the nightly build yet.  I would mark that as allowed to fail in the build matrix, so that we can monitor it as the official release of 3.6 gets closer.

abingham:
Should I go ahead and merge this in even with the travis failures? (I'm sorry, but I'm too pressed for time right now to come to my own conclusions!)

atodorov:
@abingham please merge it. Only Pythin nightly job fails, the rest are OK.

abingham:
Looks like it needs a merge.

atodorov:
ready to merge

abingham:
Thanks!

PR #166, opened 2016-09-08
Fix some code smells & update style

atodorov:
Fix some issues identified by Landscape.io, mostly cosmetic.

### DIFF
--- a/cosmic_ray/cli.py
+++ b/cosmic_ray/cli.py
@@ -348,7 +348,7 @@ def main(argv=None):
 
     command = configuration['<command>']
     if command is None:
-        command == 'help'
+        command = 'help'
 
     try:
         handler = COMMAND_HANDLER_MAP[command]
--- a/cosmic_ray/cli.py
+++ b/cosmic_ray/cli.py
@@ -56,7 +56,7 @@ Get the top-level help, or help for <command> if specified.
     if not command:
         options = OPTIONS
     elif command not in COMMAND_HANDLER_MAP:
-        LOG.error('"{}" is not a valid cosmic-ray command'.format(command))
+        LOG.error('"%s" is not a valid cosmic-ray command', command)
         options = OPTIONS
     else:
         options = COMMAND_HANDLER_MAP[command].__doc__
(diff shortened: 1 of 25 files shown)


cosmic_ray/cli.py
```diff
@@ -56,7 +56,7 @@ def handle_help(config):
     if not command:
         options = OPTIONS
     elif command not in COMMAND_HANDLER_MAP:
-        LOG.error('"{}" is not a valid cosmic-ray command'.format(command))
+        LOG.error('"%s" is not a valid cosmic-ray command' % command)
```
abingham: I thought the best practice here was `LOG.error('"%s' is not a valid cosmic-ray command', command)`, i.e. to pass interpolation values to `error()` rather than filling in template values ourselves. This way the interpolation never happens unless the logger actually needs to generate the display value.

abingham:
Except for my question about string interpolation for log messages, this looks good.

cosmic_ray/cli.py
atodorov: Well I also though like you did but the tool says otherwise. I don't have a preference really. If you don't like this particular optimization I will remove it and leave the rest, just let me know what to do.

abingham: Is this the warning you're seeing?

> `W:102,10: Use % formatting in logging functions and pass the % parameters as arguments (logging-format-interpolation)`

If so, I'm certain that they're asking us to pass the interpolation arguments as arguments to the logging function, and not to do string interpolation with `% or `format()`. Passing them as arguments is certainly a best practice, and it looks like the [pylint pull request for this warning](https://bitbucket.org/logilab/pylint/pull-requests/169/add-new-warning-logging-format/diff) was intended to get people to use the argument-passing form. 

I agree, though, that the language of the warning is a bit confusing. It really does sound like they want you to pass the arguments using `%`.

atodorov:
@abingham - thanks for actually finding the pylint error message, I've updated the PR and also rebased to master. Here's the current status:
https://landscape.io/github/MrSenko/cosmic-ray/96

Maybe you can enable Landscape.io integration for the official repo. Thoughts ?

abingham:
Yeah, landscape for the main repo is a good idea! I'll see if I can set it up this morning.

PR #171, opened 2016-09-13
Prevent modifying session.json when doing cosmic-ray report

atodorov:
this is fixed by requiring TinyDB >= 3.2.1. The particular commit
appears to be https://github.com/msiemens/tinydb/commit/331bb0.

@msiemens would know for sure b/c this is not mentioned in their changelog and the changelog got a bit messed up in the last version.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -69,7 +69,7 @@ INSTALL_REQUIRES = [
     'pathlib',
     'pytest',
     'stevedore',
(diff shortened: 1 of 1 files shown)


msiemens:
Oops, sorry for the incomplete changelog! Yes, this was fixed in msiemens/tinydb#104 and published as v3.2.1.

Issue #170, opened 2016-09-13
AND/OR replacement notes - consecutive operations with the same operator produce a single mutation


atodorov:
Consider the expression `condA and condB` which is mutated into `condA OR condB` - this is as expected. Now consider a more complicated expression: `condA and condB and condC`. You may expect that there will be two mutations here:
`condA OR condB and condC` and 
`condA and condB OR condC`, a bonus mutation may be considered as well
`condA OR condB OR condC`

However this is not the case. Only the last mutation is produced!  Consecutive operations with the same operator, such as `a or b or c`, are collapsed into one node with several values. See the example below
```
>>> ast.dump(ast.parse('meat and bread'))
"Module(body=[Expr(value=BoolOp(op=And(), values=[Name(id='meat', ctx=Load()), Name(id='bread', ctx=Load())]))])"
>>> 
>>> ast.dump(ast.parse('meat and bread and mayo'))
"Module(body=[Expr(value=BoolOp(op=And(), values=[Name(id='meat', ctx=Load()), Name(id='bread', ctx=Load()), Name(id='mayo', ctx=Load())]))])"
>>> 
>>> ast.dump(ast.parse('meat and bread and mayo and eggs'))
"Module(body=[Expr(value=BoolOp(op=And(), values=[Name(id='meat', ctx=Load()), Name(id='bread', ctx=Load()), Name(id='mayo', ctx=Load()), Name(id='eggs', ctx=Load())]))])"
```

@abingham this is documented at http://greentreesnakes.readthedocs.io/en/latest/nodes.html#BoolOp but I've missed it somehow. Where do you think is the proper place in Cosmic Ray to add a warning about this behavior ?

abingham:
Interesting. With the current design, we probably need to update our operators and mutating core a bit to account for nodes like this. Roughly, I'm picturing something like this:

1. `Operator.visit_mutation_site()` needs to accept a new `count` argument indicating how many mutations can be made by an operator on a particular node. It will forward this new parameter along to the cores' `visit_mutation_site()`s (which also need to be updated).
2. The `Visit_*()` methods on operators need to pass this new `count` parameter when they call `visit_mutation_site()`. So in this case the operator will need to see how long the `values` field of the `Expr` node is and use that to determine `count`.
3. `Operator.mutate()` will accept a new `count` argument indicating which of potentially multiple mutations to make.
[...]

PR #172, opened 2016-09-20
MutationCore and mutate() updates

atodorov:
These are the updates for #170.

### DIFF
--- a/cosmic_ray/mutating.py
+++ b/cosmic_ray/mutating.py
@@ -49,7 +49,7 @@ class MutatingCore:
         """
         return self._activation_record
 
-    def visit_mutation_site(self, node, op):
+    def visit_mutation_site(self, node, op, num_mutations):
         """Potentially mutate `node`, returning the mutated version.
 
         `Operator` calls this when AST iteration reaches a
@@ -57,18 +57,22 @@ class MutatingCore:
         mutation, the subclass instance will be asked to perform the
         mutation.
         """
-        if self._count == self._target:
-            self._activation_record = {
-                'operator': _full_module_name(op),
-                'occurrence': self._target,
-                'line_number': cosmic_ray.util.get_line_number(node)
-            }
+        for idx in range(num_mutations):
+            if self._count == self._target:
+                self._activation_record = {
+                    'operator': _full_module_name(op),
+                    'occurrence': self._target,
+                    'line_number': cosmic_ray.util.get_line_number(node)
+                }
 
-            old_node = node
-            node = op.mutate(old_node)
-            ast.copy_location(node, old_node)
+                old_node = node
+                node = op.mutate(old_node, idx)
+                ast.copy_location(node, old_node)
+                self._count += 1
+                break
+            else:
+                self._count += 1
 
-        self._count += 1
         return node
 
     def repr_args(self):
--- a/cosmic_ray/mutating.py
+++ b/cosmic_ray/mutating.py
@@ -67,7 +67,8 @@ class MutatingCore:
 
                 old_node = node
                 node = op.mutate(old_node, idx)
-                ast.copy_location(node, old_node)
+                # add lineno and col_offset for newly creted nodes
+                ast.fix_missing_locations(node)
                 self._count += 1
                 break
             else:
--- a/cosmic_ray/operators/boolean_replacer.py
+++ b/cosmic_ray/operators/boolean_replacer.py
@@ -32,15 +32,39 @@ class ReplaceTrueFalse(Operator):
 
 class ReplaceAndWithOr(Operator):
     """An operator that swaps 'and' with 'or'."""
-    def visit_And(self, node):  # noqa
+    def visit_BoolOp(self, node):  # noqa
         """
-            http://greentreesnakes.readthedocs.io/en/latest/nodes.html#And
+            http://greentreesnakes.readthedocs.io/en/latest/nodes.html#BoolOp
         """
-        return self.visit_mutation_site(node)
+        if isinstance(node.op, ast.And):
+            return self.visit_mutation_site(node, len(node.values))
+        else:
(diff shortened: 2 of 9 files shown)


atodorov:
Note: I will have to update the `AddNot` operator as well and check how good this works on more complex expressions but it looks like it should work just fine, mutating the individual AND or OR bits of the expression one by one until we've exhausted all possibilities.

cosmic_ray/mutating.py
```diff
@@ -49,26 +49,30 @@ def activation_record(self):
         """
         return self._activation_record
 
-    def visit_mutation_site(self, node, op):
+    def visit_mutation_site(self, node, op, count):
```
abingham: The parameter name `count` is a bit confusing since it collides with with `self._count`; we're using the same word to mean two different things here. I think it would be better to call it `num_mutations` or something more explicit.

atodorov:
@abingham I've decided to work on the Not mutation operators in a separate PR so this is ready to merge if you approve the changes. The failure in Travis is from missing stevedore for Python nightly.

cosmic_ray/mutating.py
```diff
@@ -49,26 +49,30 @@ def activation_record(self):
...
-                'occurrence': self._target,
-                'line_number': cosmic_ray.util.get_line_number(node)
-            }
-
-            old_node = node
-            node = op.mutate(old_node)
-            ast.copy_location(node, old_node)
-
-        self._count += 1
+        for idx in range(count):
```
abingham: This looks like it works fine, but the logic feels a bit heavy to me. What about something like:
```
span = self._target - self._count
if span in range(count):
    self._activation_record = ...
self._count += 1
return node
```

cosmic_ray/test/test_operators.py
```diff
@@ -86,7 +86,7 @@ def test_activation_record_created(operator, code):
 @pytest.mark.parametrize('operator,code', OPERATOR_SAMPLES)
 def test_no_activation_record_created(operator, code):
     node = ast.parse(code)
-    core = MutatingCore(1)
+    core = MutatingCore(-1)
```
abingham: Is this `-1` necessary due to the new logic in `MutatingCore`?

abingham:
Your new implementation raises an interesting prospect! Right now we use some relatively obscure techniques to generate the large family of relational-operator replacement operators. But if operators are now allowed to perform multiple mutations per node, perhaps those replacement operators can just do multiple mutations (i.e. one for each other relational operator) rather than dynamically generating a forest of classes. This new approach might also give us a more natural way to express exceptios (i.e. to replace `relational_operator_replacement.SKIP`).

cosmic_ray/operators/boolean_replacer.py
```diff
@@ -32,28 +32,68 @@ def mutate(self, node):
...
+        # replace all occurences of And()
+        # A and B and C -> A or B or C
+        node.op = ast.Or()
+
+        # or replace an operator somewhere in the middle
+        # of the expression
+        if idx and len(node.values) > 2:
+            left = node.values[:idx]
+            if len(left) > 1:
+                left = [ast.BoolOp(op=ast.And(), values=left)]
```
abingham: Regarding the problems with missing `lineno` and `col_offset` attributes: do we need to make sure that these new `BoolOp` instances have line numbers and columns? Do we need to call `ast.copy_location()` for them?

cosmic_ray/test/test_operators.py
atodorov: Yes. In this test you are looking for a mutation/node which doesn't exist but value of 1 now hits a valid mutation.

cosmic_ray/operators/boolean_replacer.py
atodorov: No. They are fixed later by ast,fix_missing_location`

cosmic_ray/mutating.py
atodorov: I'm not very confident in making this change b/c I don't understand this piece of code quite well, hence the above implementation. What puzzles me is how exactly we get to the point of calling this method ? As far as I understand it `self._target` is the N-th occurrence of a particular operator we want to mutate. After the changes the meaning of N is no longer the N-th occurrence but the N-th possible mutation for this operator. Where N ranges from 0 to num_operator * num_variations (you get what I mean). 

Even if I do the change you request above I will need the `idx` variable to pass to `mutate()` so it knows which particular variation to produce. I still need to loop over the range to get `idx`.

cosmic_ray/mutating.py
atodorov: updated

atodorov:
> Your new implementation raises an interesting prospect! Right now we use some relatively obscure techniques to generate the large family of relational-operator replacement operators. But if operators are now allowed to perform multiple mutations per node, perhaps those replacement operators can just do multiple mutations (i.e. one for each other relational operator) rather than dynamically generating a forest of classes. This new approach might also give us a more natural way to express exceptions (i.e. to replace relational_operator_replacement.SKIP).

Let's discuss this in a separate issue, definitely sounds like an improvement.

cosmic_ray/mutating.py
abingham: Yeah, it is getting a bit complex! When operators were only allowed to make one mutation per node, `self._target` represented the N-th time an operator encountered a node that it could mutate. Now `self._target` represents the N-th time an operator could perform a mutation, taking into account that an operator might make multiple mutations per node.

Similarly, `self._count` now represents how many mutations an operator has so far claimed it could make while walking the AST.

So the basic idea of the change is to say this: if the operator claims that it performs enough mutations on the current node to reach the target, then let it do it the mutation. 

Maybe the thing to do is to get your version working and stable. Then we can clean up the code later.

PR #175, opened 2016-10-05
Refactoring of MutationCore.visit_mutation_site() and comparison operators

atodorov:
@abingham these are the changes we last discussed. However I'm getting a failure result on adam tests. More precisely:

```
     def constraint(x, y):
-        return ((x == y) ^ (x != y))
+        return ((x > y) ^ (x <= y))

     def constraint(x, y):
-        return ((x == y) ^ (x != y))
+        return ((x is not y) ^ (x is y))

     def constraint(x, y):
-        return ((x == y) ^ (x != y))
+        return ((x != y) ^ (x == y))
```

I like the binary XOR mojo but that doesn't play well in the above case because we get equivalent mutations and there is nothing to kill them. Two thing here:

1) Do you have any advice/preference on how to update the tests? 
2) Do you have any idea why this test was not producing equivalent mutants previously ?


atodorov:
> 2) Do you have any idea why this test was not producing equivalent mutants previously ?

just a note as far as I can tell from the full log the failing mutations were not generated with the master branch, which is strange. We may have uncovered some hidden bug.

cosmic_ray/mutating.py
```diff
@@ -57,23 +57,25 @@ def visit_mutation_site(self, node, op, num_mutations):
...
-            else:
-                self._count += 1
-
+        # Calculate how many mutations are left before we reach the target
+        # mutation
+        span = self._target - self._count
+
+        # If the current operator will do at least that many mutations,
+        # then let it make the mutation now.
+        if self._target >= 0 and span < num_mutations:
```
abingham: Maybe I'm being overly sensitive, but this check for `self._target >= 0` really bugs me. As far as I can tell, it's only there to account for some tests, and it looks like the tests could be updated relatively easily to not require this fix. Am I missing something?

abingham:
Those failures are strange. It looks like two mutations are being made at the same time, and this should not happen. For example, this:
```
     def constraint(x, y):
-        return ((x == y) ^ (x != y))
+        return ((x > y) ^ (x <= y))
```
should be two separate mutations. But perhaps I'm misinterpreting things. I'll try running your branch on my machine and see what happens.

abingham:
FWIW, I'm seeing the same failures that you're reporting.

abingham:
I think I see what's going on. Basically, as the `operator.visit()` call traverses the AST, it doesn't stop once we've made a mutation. So, the first time a multi-mutation operator is asked to make a mutation, it does so and return the mutated node. But, the traversal continues and the `MutatingCore` once again asks for mutation of another node, which the operator performs. The bug has something to do with allowing the system to make more than one mutation in a traversal; I'm too tired (4am start with travel today) to come up with a coherent solution that I'm happy with.

 Here's some code I used to test this out:
```
import ast
from cosmic_ray.mutating import MutatingCore
from cosmic_ray.operators.comparison_operator_replacement import MutateComparisonOperator

code = "((x is not y) ^ (x is y))"
node = ast.parse(code)
print()
print(ast.dump(node))
core = MutatingCore(0)
operator = MutateComparisonOperator(core)
new_node = operator.visit(node)
print()
print(ast.dump(new_node))
```

If you instrument `MutateComparisonOperator.mutate()` to print when it runs, you'll see that this code causes it to mutate twice on two different nodes. This is the core of the problem, I think.

abingham:
I think I've got this sorted out. It's relatively simple, actually. I was just doing the wrong math initially. Here's how `MutatingCore.visit_mutation_site()` should look:

```
def visit_mutation_site(self, node, op, num_mutations):
        """Potentially mutate `node`, returning the mutated version.

        `Operator` calls this when AST iteration reaches a
        potential mutation site. If that site is scheduled for
        mutation, the subclass instance will be asked to perform the
        mutation.
        """
        # If the current operator will do at least that many mutations,
        # then let it make the mutation now.
        if self._count <= self._target < self._count + num_mutations:
            assert self._activation_record is None

            self._activation_record = {
                'operator': _full_module_name(op),
                'occurrence': self._target,
                'line_number': cosmic_ray.util.get_line_number(node)
            }

            old_node = node
            assert self._target - self._count < num_mutations
            node = op.mutate(old_node, self._target - self._count)
            # add lineno and col_offset for newly created nodes
            ast.fix_missing_locations(node)

        self._count += num_mutations
        return node
```

The biggest change is `self._count += num_mutations`. This is basically ensuring that an operator only makes one change per traversal; it increments `_count` by the number of mutations that the operator could make at this node, which is what the algorithm is really relying on.

Another critical change is in `MutateComparisonOperator.visit_Compare()`:
```
def visit_Compare(self, node):
        """
            http://greentreesnakes.readthedocs.io/en/latest/nodes.html#Compare
        """
        return self.visit_mutation_site(node, (len(node.ops) - 1) * len(self.OPERATORS))
```

The important part is the `len(node.ops) - 1)`. Without it, we end up claiming that we can make mutations from an operator to itself, i.e. we include the identity transformation in the count of mutations the operator claims. Since this makes no change by definition, we end up with a surviving (equivalent) mutant.

I've tested this a bit, but you should definitely double-check my logic.

abingham:
Thinking about this some more, we need to be careful about *any* exceptions we make for operator replacement. For example, you have:
```python
elif isinstance(from_op, ast.Eq) and to_op is ast.Is:
            pass
```
which would again result in an equivalent mutant.

atodorov:
@abingham 

looks like your logic in `MutatingCore.visit_mutation_site()` is correct, however the changes you propose for the operator mutations are not so simple:

```
def visit_Compare(self, node):
        """
            http://greentreesnakes.readthedocs.io/en/latest/nodes.html#Compare
        """
        return self.visit_mutation_site(node, (len(node.ops) - 1) * len(self.OPERATORS))
```

1) I think this needs to be `len(self.OPERATORS) -1`, not `len(node.ops) -1`
2) Even with 1) we don't solve the problem of excluding mutations to the same operator we came from. We need to exclude the from operator from the list of possible mutation operators
3) `if isinstance(from_op, ast.Eq) and to_op is ast.Is: pass` - you are correct, this will produce equivalent mutations (in fact there will be no change in the code and the test suite will not fail causing a survivor to be reported. One possibility is to move this exclusion logic in `mutate()` and adjust the operators list accordingly. Another possibility is to compare the AST before and after mutation and silently skip the run, reporting a KILLED result.

abingham:
> 1) I think this needs to be len(self.OPERATORS) -1, not len(node.ops) -1

Ah! You are correct, of course :)

> 2) Even with 1) we don't solve the problem of excluding mutations to the same operator we came from. We need to exclude the from operator from the list of possible mutation operators

Hmmm, right, I didn't think deeply enough about this. It seems like we need to do something like:

1. In `visit_Compare()` calculate the correct number of mutations based on the operators we see in `node.ops`. Because of exceptions like `==` to `is`, this count will vary from operator to operator.
2. In `mutate()` we need to look up the right from and to operators using the same sequence that (1) used in its calculations.

I can take a swing at this if you want, but if you're in the zone then go for it.

abingham:
[Here's a candidate](https://gist.github.com/abingham/01728c7fbf159c8a5e1474917145879f).

atodorov:
@abingham thanks for the gist. I was thinking along the same lines but you beat me to it. PR should be ready to merge now.

abingham:
Awesome, thanks!

PR #177, opened 2016-10-18
Refactor of UnaryOp mutations

atodorov:
@abingham I've moved out one method into `util.py` and then made use of it in the `MutateUnaryOperator` class which now provides the same functionality as `ReverseUnarySub`, `ReverseUnaryAdd` and `RemoveNot`. Also I think `ReverseUnaryAdd` was never used before.

Take into account the comment about `not` providing only the removal mutation. As of now I don't think we get much value of calling '+', '-' or '~' in front of objects or expressions where 'not' is most likely used. If my statement turns out to be invalid then the fix is 1 line removal.

Waiting for your comments.


cosmic_ray/util.py
```diff
@@ -51,3 +51,18 @@ def get_line_number(node):
         return node.lineno
     else:
         return '<UNKNOWN>'
+
+def build_mutations(ops, to_ops):
```
abingham: I think it might be helpful to add an argument `include_identity=False` which, if false, would not return an identity mutation (i.e. where `to_op` and `from_op` are the same). It seems likely that most users of this function will never want to make the identity mutation, so we could make their lives a bit simpler by not producing it.

We could make change in a separate commit, of course.

abingham:
This is a really nice change. Thanks!

PR #179, opened 2016-10-25
Update mutations for binary operators

atodorov:
@abingham my previous PR removed all binary operator mutations and this one adds them again. This time we make use of the refactored core and add a lot more possible mutations. Also notice the quick fix of the test suite. The xor in `equals` produces equivalent mutations when replaced with '+' or '-' in this context.


abingham:
This looks great! I made one comment about `build_mutations()` that we should discuss, but I'm going to merge this in now.

PR #180, opened 2016-11-03
Refactor skipping mutations to self

atodorov:
This is now handled in `build_mutations()` because it is a common operation. `_to_ops()` needs to only special case mutations depending on context, e.g. ==/is mutation.

@abingham please review

### DIFF
--- a/cosmic_ray/util.py
+++ b/cosmic_ray/util.py
@@ -64,5 +64,18 @@ def build_mutations(ops, to_ops):
     return [
         (idx, to_op)
         for idx, from_op in enumerate(ops)
-        for to_op in to_ops(from_op)
+        # note: mutations to self are excluded.
+        # None is a special mutation meaning to delete the operator!
+        # 1) when to_op is None isinstance(from_op, None) will blow up because
+        #    the second parameter needs to be a class
+        # 2) when to_op != None we do the isinstance() check to figure out
+        #    whether or not to include the operator in the list of possible mutations
+        #
+        # The `if to_op is None or isinstance(from_op, to_op)` expression handles both
+        # scenarios very elegantly. First we handle 1) and if this is True the rest of
+        # the expression is not evaluated and None is returned. Else we're in scenario 2)
+        # where the left part of the expression is False so the right part is evaluated.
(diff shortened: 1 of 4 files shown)


cosmic_ray/util.py
```diff
@@ -64,5 +64,7 @@ def build_mutations(ops, to_ops):
     return [
         (idx, to_op)
         for idx, from_op in enumerate(ops)
-        for to_op in to_ops(from_op)
+        # note: mutations to self are excluded.
+        # None is a special mutation meaning to delete the operator!
+        for to_op in to_ops(from_op) if to_op is None or not isinstance(from_op, to_op)
```
abingham: What is the `top_op is None` clause for? Is there some case where the to- and from-ops are both None?

atodorov: from_op is never None b/c this is the operator we are going to mutate. to_op is the new operator we want to use. In some cases (e.g. unary_operator_replacement.py) we also want to delete the operator, instead of replacing it with another value. A replacement of None signifies operator deletion.

abingham: yes, but why is there a special check for it here? It seems like `if to_op is None` is unnecessary, i.e. `None` would have been included in the comprehension in any case.

atodorov: There are two cases here:

1) to_op is not None so we do the isinstance() check to figure out whether or not to include the operator in the list of possible mutations
2) to_op is None. In this case isinstance(from_op, None) will blow up because the second parameter needs to be a class.

The `if to_op is None or isinstance(from_op, to_op)` expression handles both scenarios very elegantly. If 2) then the rest of the expression is not evaluated and None is returned. If 1) the left part of the expression is False so the next right part is evaluated. Since the left part of the expression has confirmed that to_op is not None then we're confident that the isinstance() method will always work.

Should I add the above explanation as a comment in the code? I though the 2 lines were sufficient but now I realize I could be more verbose.

abingham:
Ah right, sorry, it's clear now. I was reading this after a long day of
beating my head against HTML and CSS, and I wasn't in any shape to be doing
a code review ;)

PR #181, opened 2016-11-03
Fail if baseline execution fails

atodorov:
@abingham this is ready for review.

Please note the first commit in this PR. For some reason I was missing the `run_tests.sh` plumbing from #160 so I did cherry-pick of 2718e81 and adjusted for latest syntax. If you can figure out what went wrong I will be interested to hear, especially if you manage to find steps to reproduce. The only thing that comes to mind is accidentally removing this piece of code while dealing with merge conflicts on Jul 24, 2016.

### DIFF
--- a/cosmic_ray/cli.py
+++ b/cosmic_ray/cli.py
@@ -22,7 +22,7 @@ import cosmic_ray.modules
 import cosmic_ray.json_util
 import cosmic_ray.worker
 import cosmic_ray.testing
-import cosmic_ray.timing
+from cosmic_ray.timing import Timer
 from cosmic_ray.util import redirect_stdout
 from cosmic_ray.work_db import use_db, WorkDB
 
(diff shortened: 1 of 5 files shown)


atodorov:
@abingham ping

cosmic_ray/cli.py
```diff
@@ -133,10 +145,9 @@ def handle_init(configuration):
     else:
         baseline_mult = float(configuration['--baseline'])
         assert baseline_mult is not None
-        timeout = baseline_mult * cosmic_ray.timing.run_baseline(
-            configuration['--test-runner'],
-            configuration['<top-module>'],
-            configuration['<test-args>'])
+        with Timer() as t:
+            handle_baseline(configuration)
```
abingham: This is really nice!

cosmic_ray/timing.py
```diff
@@ -39,18 +37,3 @@ def __enter__(self):
 
     def __exit__(self, ex_type, ex_value, ex_traceback):
         pass
-
-
-def run_baseline(test_runner, module_name, test_args):
```
abingham: I'm a little bit concerned since the baseline is now timing one thing - the time to run the tests - while the worker is timing something else - the time to start a python process and run the tests. This is probably not a practical issue for most cases, but it's worth keeping in mind.

abingham:
Sorry for forgetting about this!

PR #186, opened 2016-12-12
Local exec

abingham:


abingham:
This still needs a bit of work.

PR #190, opened 2016-12-20
Local execution mode

abingham:
@atodorov @boxed This pull request adds support for running all of the mutation testing locally rather than via celery workers. By default, the `exec` command spawns a new local process for each mutation rather than queuing the work in celery; to use celery you pass the "--dist" flag to `exec`. 

This work is motivated by several users - and in particular @boxed - who noted that the work needed to simpy try CR out was too high, and that it should be possible to simply take CR for a spin should be minimized. Hopefully this gets us closer.

If either or both of you has time, could you look the changes over? I'm pretty sure the functionality is in pretty good shape, though you should verify that if possible. I'm more concerned that the documentation is correct and helpful, i.e. that new potential users will find it easy to try CR our quickly.

This line of work is not complete, but I think it's ready to get in front of the general public. In particular, I'm probably going to create an abstraction/plugin system for "execution modes" of which "local" and "celery" will be the first two examples. This will give us (or others) the flexibility to add other distribution modes if we want, perhaps supporting environments like server clusters or cloud systems.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,12 @@ processes they spawn - need to be able to import the modules you want to test.
 As a result, you generally want to start them in the virtual environment into
 which you've installed Cosmic Ray.
 
+Also remember that the workers need to be able to find and execute the tests as
+expressed to the `init` command. In other words, if you used `cosmic-ray init .
+. . -- tests` to initialize a session, the test loader (whether local or
+distributed) will look for tests in the `test` directory. So you need to make
+sure that the worker processes are running in directory where this makes sense.
+
 Finally, once the worker(s) are running you need to use the `--dist` flag when
 you run `cosmic-ray exec`:
 ```
--- a/README.md
+++ b/README.md
@@ -114,6 +114,6 @@ cosmic-ray exec --dist
 ```
 
 Note that all of the other Cosmic Ray commands --- `init`, `report`, etc. ---
-don't need the `--dist` flag; only `exec` uses it.
+don't need the `--dist` flag; only `exec` and `run` use it.
 
 **[Further documentation is available at readthedocs](http://cosmic-ray.readthedocs.org/en/latest/).**
--- a/README.md
+++ b/README.md
@@ -40,28 +40,9 @@ python setup.py install
 We recommend installing it into a virtual environment. Often it makes sense to
 install it into the virtual environment of the package you want to test.
 
-### Install RabbitMQ and start a worker
-
-Next you need to install [RabbitMQ](https://www.rabbitmq.com/). Cosmic Ray uses
-this message queue (via [Celery](http://www.celeryproject.org/)) to distribute
-testing tasks. Once installed, start the server. This is very platform-specific,
-so see the instructions for RabbitMQ on how to do this.
-
-Once RabbitMQ is running, you need to start one or more Cosmic Ray worker tasks
-to listen for commmands on the queue. Start a worker like this:
-
-```
-celery -A cosmic_ray.tasks.worker worker
-```
-
-You can start as many workers as you want. Be aware that these workers - and the
-processes they spawn - need to be able to import the modules you want to test.
-As a result, you generally want to start them in the virtual environment into
-which you've installed Cosmic Ray.
-
 ### Create a session and run tests
 
-Finally, you're ready to start killing mutants. Cosmic Ray uses a notion of
+Now you're ready to start killing mutants. Cosmic Ray uses a notion of
 *sessions* to encompass a full mutation testing suite. Since mutation testing
 runs can take a long time, and since you might need to stop and start them,
 sessions store data about the progress of a run. The first step in a full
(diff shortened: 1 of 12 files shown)


boxed:
Great!

README.md
```diff
@@ -93,4 +74,46 @@ This will print out a bunch of information about the work that was performed,
...
+sure that the worker processes are running in directory where this makes sense.
+
+Finally, once the worker(s) are running you need to use the `--dist` flag when
+you run `cosmic-ray exec`:
+```
+cosmic-ray exec --dist
+```
+
+Note that all of the other Cosmic Ray commands --- `init`, `report`, etc. ---
+don't need the `--dist` flag; only `exec` uses it.
```
atodorov: Maybe document that the `run` command also supports `--dist` ?

cosmic_ray/commands/execute.py
```diff
@@ -1,25 +1,59 @@
...
+            for r in results:
+                job_id, command, (result_type, result_data) = r.get()
+                yield job_id, command, result_type, result_data
+        finally:
+            if self.purge_queue:
+                cosmic_ray.tasks.celery.app.control.purge()
+
+
+def execute(work_db, dist=True):
     """Execute any pending work in `work_db`, recording the results.
```
atodorov: You have removed `purge_queue` parameter. What if I want to change it to `False`

cosmic_ray/commands/execute.py
```diff
@@ -1,25 +1,59 @@
...
-            timeout,
-            work_db.pending_work)
-
-        for r in results:
-            job_id, command, (result_type, result_data) = r.get()
-            work_db.add_results(job_id, command, result_type, result_data)
-    finally:
-        if purge_queue:
-            cosmic_ray.tasks.celery.app.control.purge()
+    executor = CeleryExecutor() if dist else local_executor
```
atodorov: Pass `purge_queue` to the constructor once introduced back

docs/index.md
```diff
@@ -139,6 +122,27 @@ database for your session.
...
+it will attempt to mutate any and all code in a module. This means that if you
+have test code in the same module as your code under test, Cosmic Ray will
+happily mutate the test code along with the production code. This is probably
+not what you want.
+
+The best way to avoid this problem is to keep your test code in separate modules
+from your production code. This way you can tell Cosmic Ray precisely what to
+mutate.
+
+Ideally, your test code will be in a different package form your production
```
atodorov: `form` -> `from`

docs/index.md
```diff
@@ -139,6 +122,27 @@ database for your session.
...
+test. This way you can tell Cosmic Ray to mutate an entire package without
```
atodorov: `test` -> `code`

test_project/run_tests.sh
```diff
@@ -1,31 +1,20 @@
...
+    RESULT=`cosmic-ray survival-rate adam_tests.$CONFIG`
+    if [ $RESULT != 0.00 ]; then
+        cosmic-ray report adam_tests.$CONFIG
+        exit 1
+    fi
+done
 
-# Run import tests
+# #
+Run import tests
```
atodorov: This is still a comment and it should stay that way, doesn't it ?

atodorov:
@abingham can you explain how the local executor works ? I think it should fork the process and execute everything separately, not sure if this is the current implementation. Thanks!

README.md
abingham: Good point. I'm actually thinking of getting rid of the run command. It really only exists because there was originally distinction between init and exec, and i'm not sure how useful it is in practice now.

cosmic_ray/commands/execute.py
abingham: My plan is to allow each execution mode to have its own configuration options, and "purge_queue" would be specific to celery. It makes no sense for local execution, and I figure most people are leaving it set to true for now, so I thought it was safe to leave out for now.

cosmic_ray/commands/execute.py
abingham: Yes.

test_project/run_tests.sh
abingham: Good catch!

abingham:
@atodorov The local executor launches a separate process for each mutation/testing run, but it does all of that launching from the top-level `cosmic-ray exec` process. I couldn't see a reason to have an intermediate process to do the launching, but definitely let me know if I've missed something important!

abingham:
@atodorov Regarding the silently failing travis tests, I've made issue #191 to look into it.

cosmic_ray/commands/execute.py
atodorov: OK, this is acceptable in such case. Maybe add a new task/TODO note so this doesn't get forgotten ?

docs/index.md
atodorov: Missed this one. I believe it should be `production code` not `production test`.

atodorov:
@abingham you may want to squash and re-arrange these commits before merge but that's up to you

abingham:
I did a local merge and push of this branch, so there's no need for this pull request any more.

PR #196, opened 2017-01-25
Fixed importing ImportError for Python 3.6 and after

grapherd:
When using Python 3.6 with cosmic-ray 1.0.0a1, found
that there is a ImportError cause by Python 3.6.

[class importlib.abc.Loader.create_module](https://docs.python.org/3/library/importlib.html#importlib.abc.Loader.create_module)
is now a must to defined in 3.6.

```python
job ID 22:Outcome.KILLED:hello
command: cosmic-ray worker hello number_replacer 0 nose -- test_hello.py
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/nose/failure.py", line 39, in runTest
    raise self.exc_val.with_traceback(self.tb)
  File "/usr/lib/python3.6/site-packages/nose/loader.py", line 417, in loadTestsFromName
    addr.filename, addr.module)
  File "/usr/lib/python3.6/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/usr/lib/python3.6/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/usr/lib/python3.6/imp.py", line 234, in load_module
    return load_source(name, filename, file)
  File "/usr/lib/python3.6/imp.py", line 172, in load_source
    module = _load(spec)
  File "<frozen importlib._bootstrap>", line 675, in _load
  File "<frozen importlib._bootstrap>", line 655, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 205, in _call_with_frames_removed
  File "//mutation-testing-in-patterns/python/example_04/test_hello.py", line 1, in <module>
    import hello
ImportError: loaders that define exec_module() must also define create_module()
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ matrix:
     - python: 3.4
     - python: 3.5
     - python: 3.5-dev
+    - python: 3.6
     - python: nightly
   fast_finish: true
   allow_failures:
--- a/cosmic_ray/importing.py
+++ b/cosmic_ray/importing.py
@@ -23,6 +23,9 @@ class ASTLoader:  # pylint:disable=old-style-class,too-few-public-methods
         self._ast = ast
         self._name = name
 
+    def create_module(self, spec):
+        return None
+
     def exec_module(self, mod):
         exec(compile(self._ast, self._name, 'exec'),  # pylint:disable=exec-used
              mod.__dict__)


abingham:
Thanks! I've been hoping to find some time to make CR work on 3.6, so I really appreciate your patch.

Does Cosmic Ray seem to work completely on 3.6 after your patch?

grapherd:
It seem work very great as I testing with `mutation-testing-in-patterns` and my test.

Maybe CR report need to parse the output if KILLED were not raising a correct exception? (or no the exception we expected)

I found this problem until I using --full-report to get this right.

abingham:
I'm not sure if there's a way to make this work. CR doesn't know what kinds of exceptions are "correct" for a test, so it assumes that any exception is the result of the mutation. I'm not sure how we'd distinguish between a) exceptions caused by the CR implementation and b) exceptions cause by the mutant. Perhaps we could examine the stack trace or something.

atodorov:
@grapherd there is no "correct exception". Whatever failure occurs during testing will kill the mutant. Also consider the case where you have an endless loop. After some time CR will kill the test process. 

The best way is to use `--full-report` and examine the output to see what's wrong. However I wonder why this error didn't present itself while executing the baseline portion of the tests. I'll see if I can investigate but will be busy most of this week.

@abingham FYI @grapherd is working on Chinese translation of a white paper I put together on RTD.

abingham:
> Chinese translation of a white paper I put together on RTD.
That's great news! I'll probably give some talks on CR this year, and I'll be sure to spread the word about the whitepaper.

> I wonder why this error didn't present itself while executing the baseline portion of the tests
I'm pretty sure that this is because we don't do any mutation during the baseline, and thus we don't use the faulty loader that @grapherd fixed.

grapherd:
@atodorov @abingham I'm not sure can we assume that if we are using operators like `binary_operators`, the result should not pop out something like `ImportError`, that we face in this bug.

If we can assume that, when we get the result from `test_runner`, we can then tag this test as `Outcome.MaybeKILLED`, not just `Outcome.KILLED`, to warn user that this may have some trouble.

---
@atodorov I'll go to FOSDEM2017 too, maybe we can meet there.

abingham:
Binary-operator replacement can result in import errors. For example:
```
if version.major > 3:
    import something
else:
    import something_else # not available on all platforms
```
This kind of thing is very common in Python. And once our operators get more complex (e.g. doing things like swallowing or passing exceptions), we'd be opening up even more routes for seeing exceptions that may or may not be from bugs in CR.

I'm somewhat sympathetic to your point of view, but I really don't want CR to get into the business of trying to guess when it might be broken. It's a complex-enough system as it is, and adding something like "maybe-killed" will primarily just serve to confuse users and developers. Instead, I'd like to keep the implementation simple and orthogonal, and then rely on testing and good practices to ensure that we're building a good system.

I will try to keep an open mind, though. If you've got good arguments for adding these kinds of "safety nets" to CR, I'm happy to hear them. But expect me to push back.

. . .

On that note, it might be possible to have a special reporting mode that looks for bugs in CR. I'd like reporting to be more plugin-based anyway, so this might be a good test case. This notional reporter could look for patterns like "ImportError" and raise a flag in it's output indicating that we may need to investigate the source of the bug.

Issue #206, opened 2017-05-03
Add exception mutations


drorasaf:
One of the common issues I've seen in code is the exception handling which tends to be non selective.
This can be verified with mutation over the exceptions which are raised.

What do you say about adding it to CR?

atodorov:
What exactly do you mean ? Changing `raise SomeException` to `raise AnotherException` ?? Can you give an example? 

I think the idea is good but I need more details so we can discuss. 

Btw check out this video. It talks about research on exception/error handling code and mentions several very common mistakes. It's been on my todo list for quite a while but I haven't gotten the time to work on it:
https://www.youtube.com/watch?v=56oNQf5oITw&t=785s&list=PLFjlI7p-h1hxBP3cIjEqePSeoBDHud5Db&index=55

My initial idea was to add something in pylint which will detect above mentioned patterns but maybe we can implement some of there into CR.

abingham:
As @atodorov said, it sounds to me like you're suggesting a mutation operator which modifies which exceptions are thrown in the mutated code. I think that could be a very useful class of mutation operators! Are you willing to try implementing it? 

Creating new mutation operators should be relatively simple with our plugin system. In fact, you could do it in a separate package to start with if you want. We can work with you on it to make sure you don't get hung up on too many snags.

drorasaf:
@atodorov , The idea is to mutate with the following options:
1. `raise ChildException` to `raise ParentException`
2. `raise SomeException` to `raise TotallyAnotherException`
3. switch between:
`raise SomeException
....
 raise AnotherException`
 to change into:
`raise AnotherException 
....
raise SomeException`

I think pylint is nice, but I'm not sure it can detect these kind of issues.

@abingham ,I will do as you suggested I will try to implement it myself in another repo. Thanks!

atodorov:
@drorasaf a few notes/comments

* `raise ChildException` to `raise ParentException` - you will have to traverse the inheritance tree and produce a mutation for every parent class until (including) you reach `Exception`. There are examples in the code how to generate multiple mutations for a single operator.
* `raise SomeException` to `raise TotallyAnotherException` - I think `TotallyAnotherException` must be something that Cosmic Ray provides so that we make sure we are indeed raising a completely different exception type. However if we go this approach you have to figure out how to inject the `CosmicRayMutationPlaceholderException` into the address space of the mutated modules. I'm not sure how to do this, maybe we only need to import it above all the AST modifications code? @abingham any ideas ?
```
[...]

PR #209, opened 2017-05-15
add exception swallow operator

drorasaf:
I tried to find but couldn't what are the coding conventions for the repository.

### DIFF
--- a/cosmic_ray/operators/exception_swallow.py
+++ b/cosmic_ray/operators/exception_swallow.py
@@ -1,0 +1,20 @@
+import ast
+
+from .operator import Operator
+
+
+class ExceptionSwallow(Operator):
+
+    """An operator that deleter exception handling."""
+
+    def visit_ExceptHandler(self, node):  # noq
+        """ No need to replace Pass with itself, it is not a mutant """
+        if len(node.body) == 1 and isinstance(node.body[0], ast.Pass):
+            return node
+        return self.visit_mutation_site(node)
+
+    def mutate(self, node, _):
+        """Remove the handler, create an empty one"""
+        new_node = ast.ExceptHandler(type=node.type, name=node.name,
+                                     body=[ast.Pass()])
+        return new_node
--- a/setup.py
+++ b/setup.py
@@ -52,6 +52,9 @@ operators = [
 
     'break_continue_replacement ='
     'cosmic_ray.operators.break_continue:ReplaceBreakWithContinue',
+
+    'exception_swallow ='
+    'cosmic_ray.operators.exception_swallow:ExceptionSwallow',
 ]
 
 INSTALL_REQUIRES = [
--- a/test/test_operators.py
+++ b/test/test_operators.py
@@ -18,6 +18,7 @@ from cosmic_ray.operators.boolean_replacer import (ReplaceTrueFalse,
 from cosmic_ray.operators.break_continue import (ReplaceBreakWithContinue,
                                                  ReplaceContinueWithBreak)
 from cosmic_ray.operators.number_replacer import NumberReplacer
+from cosmic_ray.operators.exception_swallow import ExceptionSwallow
 from cosmic_ray.mutating import MutatingCore
 
 
@@ -54,6 +55,7 @@ OPERATOR_SAMPLES = [
     (AddNot, 'A if B else C'),
     (AddNot, 'assert isinstance(node, ast.Break)'),
     (AddNot, 'while True: pass'),
+    (ExceptionSwallow, 'try: x = 1 \nexcept FileExistsError: x=2 \nexcept IOError: pass'),
     (ReplaceBreakWithContinue, 'while True: break'),
     (ReplaceContinueWithBreak, 'while False: continue'),
     (NumberReplacer, 'x = 1'),
--- a/test_project/adam.py
+++ b/test_project/adam.py
@@ -91,3 +91,15 @@ def trigger_infinite_loop():
     # when `while object()` becomes `while not object()`
     # the code below will be triggered
     return result
+
+
+def swallow_exception():
+    result = None
+
+    try:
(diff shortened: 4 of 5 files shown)


drorasaf:
#206

abingham:
> I tried to find but couldn't what are the coding conventions for the repository.

There's nothing official. If you follow PEP8 you'll be in good shape. Also, try to follow what you see in the rest of the code.

abingham:
I don't think this is what is meant by "exception swallowing" in general. I know what you've submitted closely matches what mutpy does, but I think that model is wrong. In my understanding, exception swallowing should add a `try...exception` around a body of code which prevents any exceptions from escaping. 

What your operator does is effectively delete an arbitrary body of code in an exception handler. This is problematic because it makes it very hard to determine *why* a mutant is killed.

Also, the behavior of your operator doesn't really match its name. It's not (necessarily) swallowing any exception; it's just gutting their handlers.

drorasaf:
My idea was to take the exception handler and remove the code handling inside of it, which is similar to the situation where we take `try: expr1 exception: expr2`, and try to verify that `expr2` has significance in tests.

If you think it is better I can make the operator work for the following use case:
`try: expr1 exception: expr2 exception: expr3` -->`try: expr1 exception: expr3

abingham:
Right, I see the intent of your operator, but I'm not convinced that it's the right thing to do. First, even if this is a mutation that we want, it should not be called "exception swallow" because it doesn't do that. Rather, it removes the content of an exception handler. This isn't quite the same as "exception handler deletion" because the handler is still there, it just doesn't do anything. Maybe it should be called "exception handler clearing".

Second, I'm not convinced that the mutation is helpful. Ultimately, it replaces an unbounded amount of code with `pass`, giving the user very little information about *why* the mutation causes a failure (or success, for that matter). Most mutations are small in nature, and this one just feels "too big". I feel like its overshadowed by the mutation that will happen anyway to the code in the `except` block, ie any benefit from this operator may already be found in existing operators.

I am open to being convinced that this operator is worth including, so by all means see if there's a good motivating example. Maybe there's a reference in some mutation testing literature, or maybe the mutpy author can provide some details (since s/he seems to do what you're doing). But I'm largely convinced that your operator needs a new name.

Also, I hope I'm not discouraging you too much. I sincerely appreciate your work on this, and I look forward to eventually integrating your work into the project.

drorasaf:
The main motivation for mutation testing is to provide stability measure for the unit tests of a given project. Since one of the main issues in code is error handling, the basic goal to verify exception handling does matter.

I agree that I can change the implementation, even if I change it to `try expr1 except: expr2` --> `expr1`, the mutation will be quite big as well.

Don't worry, I'm not discouraged so easily, I have a pipeline of operators;), if you have any new ideas I will be happy to hear.
Current pipeline:
1. Exception handler deletion - causing exception propagation
2. Change Except to another
3. remove decorator from function- excluding static and class decorators,
4. remove decorator from class
5. zero iteration loop - For/While

abingham:
> Don't worry, I'm not discouraged so easily

That's good :) The work you've done so far is great, and I'm really glad to have capable contributors.

> Since one of the main issues in code is error handling, the basic goal to verify exception handling does matter.

I completely agree, and I fully support operators that mutate error-handling code in useful ways. But I'm not (yet) sure that removing an `except-block`'s body is very useful. 

I try to look at it from the user's point of view. Suppose I have 100 lines of code in some `except-block` and CR replaces it with `pass`. If mutant survives I don't know very much. Was the exception never thrown? Which parts of the removed code weren't properly tested? Was it the removal of certain interoperating lines that mattered? Who knows. That's why I think that other mutation operators working in the removed body tell us more.

Similarly, if the mutant is killed, I don't know a lot either. Generally, I would expect *any* reasonable set of tests to fail if an arbitrary 100 line code-block were removed from my code, so I'm hesitant to add operators that do that.

abingham:
> I have a pipeline of operators

All of those sound good!

drorasaf:
> Btw check out this video. It talks about research on exception/error handling code and mentions several very common mistakes. It's been on my todo list for quite a while but I haven't gotten the time to work on it:
https://www.youtube.com/watch?v=56oNQf5oITw&t=785s&list=PLFjlI7p-h1hxBP3cIjEqePSeoBDHud5Db&index=55
> 
As @atodorov  mentioned, this provides a very good motivation, since many times the exception handling is either non existent or incorrect. Usually exception handling should not consist of 100 lines of code, especially in python;)

These projects had major bugs which were the result of trivial error handling issues.

abingham:
> this provides a very strong motivation

I don't disagree that defects in error handling are both common and serious. What I'm concerned about is introducing operators that don't add to CRs ability to detect inadequate test suites (which is all that CR really does). 

> Usually exception handling should not consist of 100 lines of code, especially in python;)

If you're starting from the (correct) position that "many times the exception handling is either non existent or incorrect", you need to also accept that many programmers will happily create except block like that. I chose 100 lines arbitrarily, but the point is that we can't assume that CR is running over optimal, beautiful code. And in any case, what number should it be? I don't think it's any more helpful to delete 10 lines than 100.

If we're going to provide operators that remove blocks of code, why would we stop at except blocks? Why not loop bodies or function bodies? These strike me as obviously nonsensical: of *course* that would cause a test failure with any meaningful test suite. But I don't see how that's any different from removing except block bodies.

A fair criticism of my position is that I'm perfectly happy to allow an operator to remove an entire except clause, even if its body is large. This feels critically different to me, however, than removing just the body because of the core change in behavior.

drorasaf:
I think that the more we talk about it, the more I think it is better to ignore this pull request and to move to change the exception handler type, which creates minimal change with the same outcome.

abingham:
That sounds good to me. Close this out if you decide to go that way.

PR #210, opened 2017-05-18
add zero iteration operator only for for loop

drorasaf:
I am still struggling with 3 issues here:
1. 2 tests are affected: use continue and use break, since they use for.
2. When I mutate manually I see, the tests fail, when I run it as part of the load, it says the mutant is alive. How should I proceed?
3. I do not know how to handle mutant which rely on the return value of the for loop, for example:
```
for i in range(5):
    ....
return i  
```
This creates incompetent mutants, since i is now UnreferencedLocalVariable.

### DIFF
--- a/cosmic_ray/operators/zero_iteration_loop.py
+++ b/cosmic_ray/operators/zero_iteration_loop.py
@@ -12,7 +12,7 @@ class ZeroIterationLoop(Operator):
 
     def mutate(self, node, _):
         """Modify the For loop to evaluate to None"""
-        empty_list_node = ast.List(elts=[], ctx=ast.Load)
+        empty_list_node = ast.List(elts=[], ctx=ast.Load())
         new_node = ast.For(node.target, empty_list_node, node.body,
                            node.orelse)
         return new_node
--- a/cosmic_ray/operators/zero_iteration_loop.py
+++ b/cosmic_ray/operators/zero_iteration_loop.py
@@ -1,0 +1,18 @@
+import ast
+
+from .operator import Operator
+
+
+class ZeroIterationLoop(Operator):
+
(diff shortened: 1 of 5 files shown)


cosmic_ray/operators/zero_iteration_loop.py
```diff
@@ -0,0 +1,18 @@
...
+class ZeroIterationLoop(Operator):
+
+    """An operator that modifies numeric constants."""
+
+    def visit_For(self, node):  # noqa
+        return self.visit_mutation_site(node)
+
+    def mutate(self, node, _):
+        """Modify the For loop to evaluate to None"""
+        empty_list_node = ast.List(elts=[], ctx=ast.Load)
```
abingham: I think this needs to be "ctx=ast.Load()", i.e. an instance of `Load`, not just the class object.

abingham:
> 2 tests are affected: use continue and use break, since they use for.

I think this is OK. As long as our test suite fails when these are modified, then everything should work.

> When I mutate manually I see, the tests fail, when I run it as part of the load, it says the mutant is alive. How should I proceed?

See my inline comment regarding `ast.Load`. I think this is the source of the problem.
 
> I do not know how to handle mutant which rely on the return value of the for loop, for example:

I think an incompetent mutant in this case is the correct thing. After all, we're trying to ensure that the test suite will detect if the user accidentally miscoded the loop to have zero iterations. Failure because of an attribute error is great!

abingham:
With the exception of the `ast.Load` issue, I think this looks really good. Thanks!

drorasaf:
I fixed the error you found, thanks!

abingham:
Great work, thanks!

PR #212, opened 2017-05-19
Exception changer

drorasaf:
As discussed in #206  and in the incomplete pull request.
Created a mutator that takes one exception handler and replace it with another.

Current issue:
report phase crashes since diff in the output is null - created another issue to address this

### DIFF
--- a/cosmic_ray/operators/exception_replacer.py
+++ b/cosmic_ray/operators/exception_replacer.py
@@ -1,0 +1,20 @@
+import ast
+
+from .operator import Operator
+
+
+class ExceptionReplacer(Operator):
+
+    """An operator that modifies exception handlers."""
+
+    def visit_ExceptHandler(self, node):  # noqa
+        return self.visit_mutation_site(node)
+
+    def mutate(self, node, _):
+        """Modify the exception handler with another exception type."""
+        except_id = 'TypeError' if node.name.id == 'SystemError' else \
+            'SystemError'
+        except_type = ast.Name(id=except_id, ctx=ast.Load())
+        new_node = ast.ExceptHandler(type=except_type, name=node.name,
+                                     body=node.body)
+        return new_node
--- a/cosmic_ray/operators/exception_replacer.py
+++ b/cosmic_ray/operators/exception_replacer.py
@@ -4,10 +4,10 @@ import builtins
 from .operator import Operator
 
 
-class OutOfNoWhereException(Exception):
+class CosmicRayTestingException(Exception):
     pass
 
-setattr(builtins, OutOfNoWhereException.__name__, OutOfNoWhereException)
+setattr(builtins, CosmicRayTestingException.__name__, CosmicRayTestingException)
 
 
 class ExceptionReplacer(Operator):
@@ -19,7 +19,7 @@ class ExceptionReplacer(Operator):
 
     def mutate(self, node, _):
         """Modify the exception handler with another exception type."""
-        except_id = OutOfNoWhereException.__name__
+        except_id = CosmicRayTestingException.__name__
         except_type = ast.Name(id=except_id, ctx=ast.Load())
         new_node = ast.ExceptHandler(type=except_type, name=node.name,
                                      body=node.body)
--- a/cosmic_ray/operators/exception_replacer.py
+++ b/cosmic_ray/operators/exception_replacer.py
@@ -1,9 +1,13 @@
 import ast
+import builtins
 
 from .operator import Operator
 
+
 class OutOfNoWhereException(Exception):
     pass
+
+setattr(builtins, OutOfNoWhereException.__name__, OutOfNoWhereException)
 
 
 class ExceptionReplacer(Operator):
--- a/cosmic_ray/operators/exception_replacer.py
+++ b/cosmic_ray/operators/exception_replacer.py
@@ -2,6 +2,9 @@ import ast
 
 from .operator import Operator
 
+class OutOfNoWhereException(Exception):
+    pass
+
 
 class ExceptionReplacer(Operator):
 
@@ -12,8 +15,7 @@ class ExceptionReplacer(Operator):
 
     def mutate(self, node, _):
         """Modify the exception handler with another exception type."""
-        except_id = 'TypeError' if node.name.id == 'SystemError' else \
-            'SystemError'
+        except_id = OutOfNoWhereException.__name__
         except_type = ast.Name(id=except_id, ctx=ast.Load())
         new_node = ast.ExceptHandler(type=except_type, name=node.name,
                                      body=node.body)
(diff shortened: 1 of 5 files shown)


cosmic_ray/operators/exception_replacer.py
```diff
@@ -0,0 +1,20 @@
...
+class ExceptionReplacer(Operator):
+
+    """An operator that modifies exception handlers."""
+
+    def visit_ExceptHandler(self, node):  # noqa
+        return self.visit_mutation_site(node)
+
+    def mutate(self, node, _):
+        """Modify the exception handler with another exception type."""
+        except_id = 'TypeError' if node.name.id == 'SystemError' else \
```
abingham: What if they have a previous handler for the exception that you mutate to? That seems like a case of a truly undetectable equivalent mutation.

I can think of two ways to try to address this. One would be to traverse the other handlers and pick an exception that isn't handled. This could get messy. Another is to insert our own, "private" exception type into the module namespace and then handle that. Offhand, I like this approach a bit more.

As written, your code is probably just fine for almost all users. Maybe the thing to do is to open a new issue that gives us a place to think about this.

test/test_operators.py
```diff
@@ -18,6 +18,7 @@
 from cosmic_ray.operators.break_continue import (ReplaceBreakWithContinue,
                                                  ReplaceContinueWithBreak)
 from cosmic_ray.operators.number_replacer import NumberReplacer
+from cosmic_ray.operators.exception_replacer import ExceptionReplacer
```
abingham: Put this in alphabetical order with the other imports. (Sorry, I'm a bit picky about some things).

test_project/adam.py
```diff
@@ -101,3 +101,13 @@ def single_iteration():
         result = True
 
     return result
+
+
+def raise_exception():
```
abingham: I think it would be clearer if this code just raised an exception. Then the test could look for that exception. It would make the code more concise and let it "speak for itself" without the reader needing to understand why values are returned, etc.

test_project/tests/test_adam.py
```diff
@@ -74,3 +74,6 @@ def test_trigger_infinite_loop(self):
 
     def test_single_iteration(self):
         self.assertTrue(adam.single_iteration())
+
+    def test_raise_exception(self):
```
abingham: Related to an earlier comment: I think this should just assert that the right exception is thrown.

test_project/adam.py
drorasaf: The mutator currently change the exception handler and not the raise.
Maybe it is better to change the raise to another one

abingham: No, you're absolutely right. I wasn't paying enough attention! Sorry about that.

abingham:
The main issue I see now is that you're defining `OutOfNowhereException` in such a way that it's not actually defined in the recompiled module. When you mutate the AST you change the handled exception by changing a string. Python then tries to compile this AST, but when it does so `OutOfNowhereException` isn't defined...it's only defined in your operator's module. So we end up with a `NameError`.

You can see this in the worker output when the unittest runner is used:
```
% cosmic-ray worker adam exception_replacer 0 unittest -- tests
{"data": ["Traceback (most recent call last):\n  File \"adam\", line 109, in handle_exception\nOSError\n\nDuring handling of the above exception, another exception occurred:\n\nTraceback (most recent call last):\n  File \"/Users/abingham/projects/sixty-north/cosmic-ray/test_project/tests/test_adam.py\", line 79, in test_handle_exception\n    self.assertTrue(adam.handle_exception())\n  File \"adam\", line 1, in handle_exception\nNameError: name 'OutOfNoWhereException' is not defined\n"], "test_outcome": "killed", "worker_outcome": "normal", "diff": ["--- mutation diff ---", "--- a/Users/abingham/projects/sixty-north/cosmic-ray/test_project/adam.py", "+++ b/Users/abingham/projects/sixty-north/cosmic-ray/test_project/adam.py", "@@ -74,7 +74,7 @@", "     result = None", "     try:", "         raise IOError", "-    except IOError:", "+    except OutOfNoWhereException:", "         result = True", "     return result", " "], "module": null, "operator": "cosmic_ray.operators.exception_replacer.ExceptionReplacer", "occurrence": 0, "line_number": 110, "command_line": null, "job_id": null}%
```

For your approach to work, we need to somehow bind your exception type to a name in the scope of the module containing the mutated except clause. There are two approaches I can think of off the top of my head.

First, we could modify the `__builtin__` module, inserting your exception there. This would make your exception available everywhere, much like the standard exceptions. That would look something like this:
```
setattr(__builtin__, 'OutOfNowhereException', OutOfNowehereException)
```
This may have serious side-effects we need to think about, but I think the core technique is sound.

The other option is to try to get a reference to the except-handler's containing module node in the AST. If we have that we can probably insert your exception into the module's scope. I don't have a sketch of how to do this.

I have not tried either of these approaches myself. But your PR as it stands doesn't work, so we need to find a better solution.

drorasaf:
I am trying to understand why the mutation on my end, did not provide this error :S, any thoughts?(I will double check myself)

I think that surgically inserting the exception to the module may be more pretty, but too complex in contrast of the availability of builtins.

abingham:
Were you using pytest or unittest? Pytest wasn't providing enough
diagnostics to see the actual error, though I've just now pushed up a
change that should make it more visible.

drorasaf:
I have been using pytest, thanks.

drorasaf:
I used the setattr syntax due to simplicity, I think it provides even better cover now.
 The only option in which I can think this might not be addressed is if there is a try except clause with general exception like this:
```
try:
    do_something() that raises MyError
except MyError:
except Exception:
    handle_error()
```

abingham:
I like this quite a bit! I still have a lingering fear that we're doing something against the rules by inserting something into builtins, but I can't think of why that would break anything. I'm happy to go with this approach until we learn better. It's easier to ask forgiveness than permission, after all :)

There's only one change I'd like. I think the exception name should be `CosmicRayTestingException` rather than `OutOfNowehereException`. This will make collision less likely (though it's already very unlikely anyway), and, more importantly, it will make it easier to see what's going on when people see the exception.

Otherwise, it looks great!

abingham:
> The only option in which I can think this might not be addressed is if there is a try except clause with general exception

This is an interesting question. Right now I can't think of any problems that this will cause. If we change the handle to catch our exception rather than `Exception`, then we'll effectively prevent that handler from catching anything (since we assume nobody will throw our exception).

abingham:
One more thing to think about. It might make sense to make your new exception un-instantiable. This way we'd have a strong guarantee that nobody could ever throw it. We might then name it `CosmicRayUnthrowableException` or something like that.

There are at least two ways to approach this. One is to have its constructor `assert False` or raise some other exception. The other is to use a metaclass which prevents a new object from ever being created at all. There might be an existing pattern for this.

abingham:
I like what you've got, so let's merge it in for now. If we start to develop more exception-oriented operators we might get more sophisticated with our infrastructure (e.g. uninstantiable exceptions, etc.) Thanks!

PR #215, opened 2017-05-30
add decorator remover 

drorasaf:
with the exception of class,static and abstract methods

### DIFF
--- a/cosmic_ray/operators/remove_decorator.py
+++ b/cosmic_ray/operators/remove_decorator.py
@@ -1,0 +1,23 @@
+import ast
+import random
+
+from .operator import Operator
+
+
+class RemoveDecorator(Operator):
+    """An operator that removes a single random decorator."""
+    regular_decorators = ["classmethod", "staticmethod", "abstractmethod"]
+
+    def visit_FunctionDef(self, node):  # noqa
+        if [x for x in node.decorator_list if x.id not in self.regular_decorators]:
+            return self.visit_mutation_site(node)
+        return node
+
+    def mutate(self, node, _):
+        """Modify the decorator list to include randomly one less"""
+        decorator_candidates = [x for x in node.decorator_list if x.id not in self.regular_decorators]
+        removed_decorator = random.choice(decorator_candidates)
+        modified_decorator = [x for x in node.decorator_list if x.id != removed_decorator.id]
+        node.decorator_list = modified_decorator
+        
+        return node
--- a/cosmic_ray/operators/remove_decorator.py
+++ b/cosmic_ray/operators/remove_decorator.py
@@ -1,23 +1,27 @@
 import ast
-import random
 
 from .operator import Operator
 
 
 class RemoveDecorator(Operator):
-    """An operator that removes a single random decorator."""
+    """An operator that removes each of the non standard decorators."""
     regular_decorators = ["classmethod", "staticmethod", "abstractmethod"]
 
     def visit_FunctionDef(self, node):  # noqa
-        if [x for x in node.decorator_list if x.id not in self.regular_decorators]:
-            return self.visit_mutation_site(node)
+        decorator_candidates = [x for x in node.decorator_list if \
+                                x.id not in self.regular_decorators]
+        if decorator_candidates:
+            return self.visit_mutation_site(node, len(decorator_candidates))
+
         return node
 
-    def mutate(self, node, _):
-        """Modify the decorator list to include randomly one less"""
-        decorator_candidates = [x for x in node.decorator_list if x.id not in self.regular_decorators]
-        removed_decorator = random.choice(decorator_candidates)
-        modified_decorator = [x for x in node.decorator_list if x.id != removed_decorator.id]
+    def mutate(self, node, idx):
+        """Modify the decorator list to remove one decorator at each mutation"""
+        decorator_candidates = [x for x in node.decorator_list if \
+                                x.id not in self.regular_decorators]
+        removed_decorator = decorator_candidates[idx]
+        modified_decorator = [x for x in node.decorator_list if \
+                              x.id != removed_decorator.id]
         node.decorator_list = modified_decorator
         
         return node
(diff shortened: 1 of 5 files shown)


abingham:
I like this operator in principle, but there's a problem in its implementation. Right now you randomly remove a decorator, but this means that the mutation testing runs aren't repeatable. We need to systematically remove each decorator individually.

CR has support for performing multiple independent mutations at a single AST node. The second parameter to `Operator.visit_mutation_site()` is how many mutations the operators subclass is able to make at that site. Then, when `mutate()` is called on your operator, its second argument, `idx`, will tell you which of the mutations you should perform.

So, for example, if you're mutating a node with two non-regular decorators, you'd pass 2 as the second parameter to `visit_mutation_site()`. Then you're `mutate()` would be called twice in a CR run, once with `idx` set to 0 and another with `idx` set to 1. Each `mutate()` call would then know which decorator to remove.

Look at `cosmic_ray.operators.binary_operator_replacement` for an example.

drorasaf:
I do not mind adding it, I started with selecting one randomly due to possible long execution time, I will change it later today.
I have to say, it is nice to have quick responses on the pull requests:-)

abingham:
> I have to say, it is nice to have quick responses on the pull requests:-)

I try to be as responsive as I can be. It's good to keep contributors happy!

> I started with selecting one randomly due to possible long execution time

You're right that this is an issue, but it's an issue with the fundamental idea of mutation testing. Generally speaking, I want to address performance issues like these with infrastructural and conceptual improvements (e.g. supporting concurrent mutations across an arbitrary number of machines). We should keep the operators as pure and simple as possible.

Thanks as always for your work on CR!

cosmic_ray/operators/remove_decorator.py
```diff
@@ -0,0 +1,27 @@
+import ast
+
+from .operator import Operator
+
+
+class RemoveDecorator(Operator):
+    """An operator that removes each of the non standard decorators."""
+    regular_decorators = ["classmethod", "staticmethod", "abstractmethod"]
```
abingham: This should be a `frozenset` with an all-caps name to indicate that it's a "constant":
```
REGULAR_DECORATORS = frozenset(("classmethod", "staticmethod", "abstractmethod"))
```

cosmic_ray/operators/remove_decorator.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def visit_FunctionDef(self, node):  # noqa
+        decorator_candidates = [x for x in node.decorator_list if \
```
abingham: These lines breaks are not necessary. And I think it would read better if you put the `if` on the same line as the rest of the conditional handling:
```
decorator_candidates = [x for x in node.decorator list
                                          if x.id not in self.regular_decorators]
```

cosmic_ray/operators/remove_decorator.py
```diff
@@ -0,0 +1,27 @@
...
+
+        return node
+
+    def mutate(self, node, idx):
+        """Modify the decorator list to remove one decorator at each mutation"""
+        decorator_candidates = [x for x in node.decorator_list if \
+                                x.id not in self.regular_decorators]
+        removed_decorator = decorator_candidates[idx]
+        modified_decorator = [x for x in node.decorator_list if \
+                              x.id != removed_decorator.id]
```
abingham: This seems incorrect. If there are multiple decorators with the same id, they would all be removed instead of just the specific one we're targeting.

cosmic_ray/operators/remove_decorator.py
drorasaf: I agree, but pep8 requires it to be 80 chars long, I changed it back to single line

cosmic_ray/operators/remove_decorator.py
drorasaf: agreed, at the beginning I thought that no one would put more than one decorator with the same name, but it might be the case especially if is parametrized

cosmic_ray/operators/remove_decorator.py
abingham: Sorry, my spacing was off in the code I posted. I've updated it. I definitely agree with PEP8 that we shouldn't extend beyond 80 chars.

abingham: And just so I'm clear, when I say "line break" I mean the "\" character. We definitely need to split that comprehension over multiple lines, but inside square brackets (and other matched punctuation) Python doesn't require the line continuation character.

cosmic_ray/operators/remove_decorator.py
```diff
@@ -0,0 +1,24 @@
...
+        if decorator_candidates:
+            return self.visit_mutation_site(node, len(decorator_candidates))
+
+        return node
+
+    def mutate(self, node, idx):
+        """Modify the decorator list to remove one decorator at each mutation"""
+        decorator_candidates = [x for x in node.decorator_list if x.id not in self.regular_decorators]
+        del decorator_candidates[idx]
+        modified_decorator = decorator_candidates + [x for x in node.decorator_list if x.id in self.regular_decorators]
```
abingham: The problem now is that this potentially reorders the decorators. We need to remove one decorator but leave the ordering of the others unchanged.

drorasaf:
I can do something like this:
 modified_decorator = [x for x in node.decorator_list
                                      if x != decorator_candidates[idx]

but Im not sure regarding the comparison of the nodes, would it be sufficient?

cosmic_ray/operators/remove_decorator.py
drorasaf: I can do something like this:
modified_decorator = [x for x in node.decorator_list
if x != decorator_candidates[idx]

but Im not sure regarding the comparison of the nodes, would it be sufficient?

abingham: I think something like this will work:
```python
def mutate(self, node, idx):
    """Modify the decorator list to remove one decorator at each mutation"""
    candidates = [(i, d)
                  for (i, d) in enumerate(node.decorator_list)
                  if d.id not in self.regular_decorators]
    remove_at, _ = candidates[idx]
    del node.decorator_list[remove_at]
    return node
```

**This is completely untested**,  so please verify it does (or doesn't) do the right thing.

drorasaf: The solution is always in front of you, Thanks!

abingham:
Thanks!

PR #223, opened 2017-07-16
fix incorrect handling for nodes without id attribute

drorasaf:


abingham:
Thanks! Have you encountered decorators that don't have an "id" field?

cosmic_ray/operators/remove_decorator.py
```diff
@@ -10,7 +10,7 @@ class RemoveDecorator(Operator):
 
     def visit_FunctionDef(self, node):  # noqa
         decorator_candidates = [x for x in node.decorator_list
-                                if x.id not in self.REGULAR_DECORATORS]
+                                if hasattr(x, 'id') and x.id not in self.REGULAR_DECORATORS]
```
abingham: Actually, shouldn't this be "if not hasattr(x, 'id') or x.id not in self.REGULAR_DECORATORS"? That is, we want to remove decorators which aren't in "REGULAR_DECORATORS", and this include any decorator without an "id" field.

abingham: Also, don't we need to repeat the new logic down in `mutate()`?

abingham:
I've created a patch in the [`decorators-without-id` branch]https://github.com/sixty-north/cosmic-ray/tree/decorators-without-id) that I think addresses the issues in your patch. Have a look at it and let me know if you see any problems.

drorasaf:
> Thanks! Have you encountered decorators that don't have an "id" field?
I encountered in one of my projects, I do not know which node it was, but i definitely got an exception on that part.

before accessing id I want to make sure it exists.

drorasaf:
looks like a nice upgrade for the code:)

Issue #224, opened 2017-07-17
Python version - does it work with 2.x?


doctorlove:
I get this:

Traceback (most recent call last):
  File "C:\Python27\Scripts\cosmic-ray-script.py", line 9, in <module>
    load_entry_point('cosmic-ray==0.1.2', 'console_scripts', 'cosmic-ray')()
  File "c:\python27\lib\site-packages\pkg_resources\__init__.py", line 542, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "c:\python27\lib\site-packages\pkg_resources\__init__.py", line 2569, in load_entry_point
    return ep.load()
  File "c:\python27\lib\site-packages\pkg_resources\__init__.py", line 2229, in load
    return self.resolve()
  File "c:\python27\lib\site-packages\pkg_resources\__init__.py", line 2235, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "c:\python27\lib\site-packages\cosmic_ray\app.py", line 10, in <module>
    from cosmic_ray.mutating import create_mutants, run_with_mutant
  File "c:\python27\lib\site-packages\cosmic_ray\mutating.py", line 5, in <module>
    from .importing import using_mutant
  File "c:\python27\lib\site-packages\cosmic_ray\importing.py", line 5, in <module>
    from importlib.abc import MetaPathFinder
ImportError: No module named abc

I didn't notice anything dsaying it only works with 3.x

abingham:
Currently it does not. There's some discussion of this topic in [this issue](https://github.com/sixty-north/cosmic-ray/issues/101).

doctorlove:
Cool - I'll close this. It might be nice to state it explictly in the readme/docs.

abingham:
That's a good idea. I'll add something soon. Maybe we can talk about python2 support at the next ACCU :)

boxed:
@doctorlove Mutmut supports python 2.7: https://github.com/boxed/mutmut/)

PR #230, opened 2017-09-07
Mention Python 3 more explicitly

Wilfred:
Prevent issues like #224

I know there's a badge, but I tried to run cosmic-ray on a Python 2 project without realising it's Python 3 only.


abingham:
Thanks!

PR #239, opened 2017-09-19
dump: produce valid JSON

blueyed:


abingham:
This intentionally produces invalid JSON. If we try to build a list of all work records, that could include billions of records on some projects. This would be prohibitive to save and load. As it is, we can process one record at a time.

blueyed:
I see / thought so..

Makes it harder to format it using tools like /usr/bin/core_perl/json_pp though.

Fortunately `jq` handles it.

PR #248, opened 2017-10-30
add tox to run simply all tests locally

gaborbernat:


abingham:
Thanks! Is there a way to run the end-to-end tests via tox as well? I.e. the tests in `test_project/run_tests.sh`.

PR #251, opened 2017-10-30
add style checker via tox (flake8 and pylint)

gaborbernat:
- also fixed all check fails
- ``tox -e lint`` can be used in CI, and locally to test that style is correct

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,6 +1,5 @@
 [*]
 charset=utf-8
-end_of_line=lf
 insert_final_newline=true
 indent_style=space
 indent_size=4
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+[*]
+charset=utf-8
+end_of_line=lf
+insert_final_newline=true
+indent_style=space
+indent_size=4
+max_line_length=79
--- a/cosmic_ray/commands/new_config.py
+++ b/cosmic_ray/commands/new_config.py
@@ -24,9 +24,7 @@ def new_config():
 
     Returns: A new configuration as a single string.
     """
-    conf = {}
-
-    conf['module'] = qprompt.ask_str("Top-level module")
+    conf = {'module': qprompt.ask_str("Top-level module")}
 
     menu = qprompt.Menu()
     for i, r in enumerate(cosmic_ray.plugins.test_runner_names()):
--- a/cosmic_ray/commands/new_config.py
+++ b/cosmic_ray/commands/new_config.py
@@ -27,16 +27,16 @@ def new_config():
     conf = {'module': qprompt.ask_str("Top-level module")}
 
     menu = qprompt.Menu()
-    for i, r in enumerate(cosmic_ray.plugins.test_runner_names()):
-        menu.add(str(i), r)
+    for at_pos, test_name in enumerate(cosmic_ray.plugins.test_runner_names()):
+        menu.add(str(at_pos), test_name)
     conf['test_runner'] = menu.show(header="Test runner",
                                     returns="desc")
 
     conf['test_args'] = qprompt.ask_str('Test args')
 
     menu = qprompt.Menu()
-    for i, e in enumerate(['local', 'celery']):
-        menu.add(str(i), e)
+    for at_pos, env in enumerate(['local', 'celery']):
+        menu.add(str(at_pos), env)
     conf['engine'] = menu.show(header="Execution engine",
                                returns="desc")
     return TEMPLATE.format(**conf)
--- a/cosmic_ray/commands/new_config.py
+++ b/cosmic_ray/commands/new_config.py
@@ -27,16 +27,17 @@ def new_config():
     conf = {'module': qprompt.ask_str("Top-level module")}
 
     menu = qprompt.Menu()
-    for at_pos, test_name in enumerate(cosmic_ray.plugins.test_runner_names()):
-        menu.add(str(at_pos), test_name)
+    test_runners = cosmic_ray.plugins.test_runner_names()
+    for at_pos, test_runner in enumerate(test_runners):
+        menu.add(str(at_pos), test_runner)
     conf['test_runner'] = menu.show(header="Test runner",
                                     returns="desc")
 
     conf['test_args'] = qprompt.ask_str('Test args')
 
     menu = qprompt.Menu()
-    for at_pos, env in enumerate(['local', 'celery']):
-        menu.add(str(at_pos), env)
+    for at_pos, engine_name in enumerate(['local', 'celery']):
+        menu.add(str(at_pos), engine_name)
     conf['engine'] = menu.show(header="Execution engine",
                                returns="desc")
     return TEMPLATE.format(**conf)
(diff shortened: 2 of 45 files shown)


cosmic_ray/modules.py
```diff
@@ -26,7 +26,7 @@ def fixup_module_name(name):
     return name
 
 
-def find_modules(name, excludes=None):
+def find_modules(_name, excludes=None):
```
abingham: Why did you make this change? I see that there's a duplication of `name` below, but you seem to have renamed them both.

cosmic_ray/config.py
```diff
@@ -12,8 +14,8 @@ def load_config(filename=None):
     if filename is None:
         return yaml.safe_load(sys.stdin)
 
-    with open(filename, mode='rt') as f:
-        return yaml.safe_load(f)
+    with open(filename, mode='rt') as handler:
```
abingham: Use `handle` instead of `handler`.

cosmic_ray/mutating.py
```diff
@@ -49,7 +47,8 @@ def activation_record(self):
         """
         return self._activation_record
 
-    def visit_mutation_site(self, node, op, num_mutations):
+    def visit_mutation_site(self, node, op,  # pylint: disable=invalid-name
```
abingham: Was pylink marking `visit_mutation_site` as an invalid name?

cosmic_ray/operators/binary_operator_replacement.py
```diff
@@ -4,28 +4,25 @@
...
+    """The sequence of operators which `from_op` could be mutated to."""
     for to_op in OPERATORS:
         yield to_op
 
 
 class MutateBinaryOperator(Operator):
     """An operator that modifies binary operators."""
 
-    def visit_BinOp(self, node):
+    def visit_BinOp(self, node):  # pylint: disable=invalid-name
```
abingham: Was this also flagged as an invalid name?

cosmic_ray/timing.py
```diff
@@ -18,6 +18,7 @@ class Timer:
     """
 
     def __init__(self):
+        self._start = None
```
abingham: This seems unnecessary.

cosmic_ray/work_db.py
```diff
@@ -149,7 +149,9 @@ def pending_work(self):
         """The sequence of pending `WorkItem`s in the session."""
         table = self._work_items
         work_item = tinydb.Query()
-        pending = table.search(work_item.worker_outcome == None)
+        out = work_item.worker_outcome
```
abingham: This whole change feels a lot more confusing than the original. Is there a reason for it?

cosmic_ray/work_db.py
```diff
@@ -169,9 +171,9 @@ def use_db(path, mode=WorkDB.Mode.create):
       FileNotFoundError: If `mode` is `Mode.open` and `path` does not
         exist.
     """
-    db = WorkDB(path, mode)
+    database = WorkDB(path, mode)
     try:
-        yield db
+        yield database
+        database.close()
```
abingham: It looks like this call to `close` should probably be in a `finally` block. I'm not sure why it was in the `except` block to begin with...

setup.py
```diff
@@ -1,16 +1,17 @@
...
+from setuptools import setup, find_packages
+
 
 def read(*names, **kwargs):
     with io.open(
         os.path.join(os.path.dirname(__file__), *names),
         encoding=kwargs.get("encoding", "utf8")
-    ) as fp:
-        return fp.read()
+    ) as handler:
```
abingham: `handle` instead of `handler`.

cosmic_ray/commands/new_config.py
```diff
@@ -24,21 +24,19 @@ def new_config():
...
     """
-    conf = {}
-
-    conf['module'] = qprompt.ask_str("Top-level module")
+    conf = {'module': qprompt.ask_str("Top-level module")}
 
     menu = qprompt.Menu()
-    for i, r in enumerate(cosmic_ray.plugins.test_runner_names()):
-        menu.add(str(i), r)
+    for at_pos, test_name in enumerate(cosmic_ray.plugins.test_runner_names()):
```
abingham: The name `test_name` is incorrect. It should be something like `test_runner` or `test_runner_name` (or just `r`).

cosmic_ray/commands/new_config.py
```diff
@@ -24,21 +24,19 @@ def new_config():
...
+        menu.add(str(at_pos), test_name)
     conf['test_runner'] = menu.show(header="Test runner",
                                     returns="desc")
 
     conf['test_args'] = qprompt.ask_str('Test args')
 
     menu = qprompt.Menu()
-    for i, e in enumerate(['local', 'celery']):
-        menu.add(str(i), e)
+    for at_pos, env in enumerate(['local', 'celery']):
```
abingham: The name `env` is incorrect. It should be `engine` or perhaps `engine_name`.

.editorconfig
```diff
@@ -0,0 +1,7 @@
+[*]
```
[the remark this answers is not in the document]
gaborbernat: it should, many editors support it to automatically pick up project configurations :-)

cosmic_ray/mutating.py
gaborbernat: it's the 'op' :-) too short for argument name according to pylint

cosmic_ray/operators/binary_operator_replacement.py
gaborbernat: BinOp is not pep-8 compliant, at best could be 'bin_op'

cosmic_ray/work_db.py
gaborbernat: otherwise we'll not fit in 80 chars :sad:

cosmic_ray/timing.py
gaborbernat: it is according to Python, but not according pylint, otherwise the linter complies that we assign to something that was not first defined in init

cosmic_ray/work_db.py
abingham: I'd much rather exceed 80 chars than use this confusing version.

.editorconfig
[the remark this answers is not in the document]
abingham: Cool! This is new to me.

gaborbernat:
@abingham now this should be good to merge 👍

cosmic_ray/work_db.py
abingham: I'd really like this changed back to how it was.

abingham:
There's just one change I'd like before this is ready. I think there may be some work to do once it's merged with the latest changes on master. 

Also, if it's not too painful, I'd like you to squash all of the changes into a single commit. This is not crucial, but it would be nice. Thanks!

cosmic_ray/work_db.py
gaborbernat: @abingham the problem is that then it will not fit within the number of columns allowed, and the linting will fail

abingham: Then the linting is broken.

abingham: What I mean by this is that, in this case, linting is asking us to do the wrong thing. Expressive, readable code is more important than obeying the rules in the linter.

PR #289, opened 2017-11-09
Regular indentation of subcommand docstrings.

rob-smallshire:
Takes advantage of the new dedent support in docopt_subcommand to allow
the subcommand docstrings to be formatted more normally.

### DIFF
--- a/cosmic_ray/cli.py
+++ b/cosmic_ray/cli.py
@@ -33,9 +33,9 @@ LOG = logging.getLogger()
 def handle_baseline(args):
     """usage: cosmic-ray baseline <config-file>
 
-Run an un-mutated baseline of the specific configuration. This is largely like
-running a "worker" process, with the difference that a baseline run doesn't
-mutate the code.
+    Run an un-mutated baseline of the specific configuration. This is largely like
+    running a "worker" process, with the difference that a baseline run doesn't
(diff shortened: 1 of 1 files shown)


abingham:
Hmmm. It looks like you're running afoul of the linter. I'm not entirely sure we should have linting in the commit process, at least not as a blocking step. It was added by a well-intentioned contributor in the sprint, but it has caused a fair amount of churn.

rob-smallshire:
Failing on a line length lint – especially at a draconian 79 characters is unhelpful. If it's trying to follow PEP 8, it's still wrong as PEP 8 says docstrings should be wrapped at 72.

I have a strong preference for code that's easy to read, rather than code which follows some alleged "rules" which are actually just guidelines. This means the guides should be followed most of the time, not all of the time, and I'm dead-set against having a line-length listing rule as a commit gate. If there is to be a hard line-length limit I would advocate for a generous limit like 120.

That said, I'll be happy to re-wrap this text to a shorter limit, especially as it won't get re-wrapped on output to the console.

abingham:
I'll remove the linting gate and make an issue to reconsider how to best do it. Some of the linting stuff is probably useful, but I agree that it's generally not useful as a commit gate.

PR #297, opened 2017-11-10
Returning error codes from cli.main().

abingham:
Also calling sys.exit() with error codes from cli's __main__-block.


rob-smallshire:
This is much clearer.

PR #298, opened 2017-11-10
Improved robustness of command line handling.

rob-smallshire:
Terse error messages instead of stack traces. More useful exit codes.

### DIFF
--- a/cosmic_ray/cli.py
+++ b/cosmic_ray/cli.py
@@ -290,6 +290,8 @@ def common_option_handler(config):
     if config['--verbose']:
         logging.basicConfig(level=logging.INFO)
 
+_SIGNAL_EXIT_CODE_BASE = 128
+
 
 def main(argv=None):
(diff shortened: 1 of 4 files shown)


rob-smallshire:
I'm not sure what I've screwed up here. The one modified and two new tests and all passing on my local build.

abingham:
@rob-smallshire In `plugins/execution-engines/celery3/setup.py`, remove `cosmic_ray` from the `install_requires` list. This dependency seems to be causing tox to install cosmic_ray from pypi before it gets installed from source.

Also, you're going to hit some style and linting errors as well. You can use `tox -e style,lint` to see these. One of these is the "invalid-name" pylint checker complaining about the variable name `e`. I'd be happy to just add a blanket exception for this in `.pylintrc`. The others are pep8 complaints about the number of blank lines before a function...I'd prefer to follow this guideline if only because it reduces editor noise.

PR #304, opened 2017-11-11
Progress reporting

rob-smallshire:
Dependent on #302 (seems to include the same changes as #302 - I guess GitHub isn't smart enough to deal with dependent PRs)

This change makes cosmic-ray respond to SIGINFO while the exec command is running by printing a short progress report to stderr.

Includes a small progress reporting framework which could also be used for the other long-running commands, such as init.

### DIFF
--- a/cosmic_ray/cli.py
+++ b/cosmic_ray/cli.py
@@ -21,6 +21,7 @@ import cosmic_ray.modules
 import cosmic_ray.plugins
 import cosmic_ray.worker
 from cosmic_ray.config import ConfigError, get_db_name, load_config
+from cosmic_ray.progress import report_progress
 from cosmic_ray.testing.test_runner import TestOutcome
 from cosmic_ray.timing import Timer
 from cosmic_ray.util import redirect_stdout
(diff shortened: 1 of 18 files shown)


abingham:
This looks good. What do you think about integrating it with `logging`? I'm imagining something like:
```python
import logging
import progress

logger = addProgressReportingHandler(logging.getLogger())

def some_func():
    logger.log(progress.PROGRESS, "starting frobincation")
    # frobnicate
    logger.log(progress.PROGRESS, "frobincation complete")
```

This could be polished a bit, but you get the idea.

rob-smallshire:
I think that's a great idea. If we installed a global progress log handler which stored the most recent log message, this message could be retrieved at some later time by the signal handler.

I the only downside I can think of for using logging is that it precludes nice "graphical" progress displays in the terminal. I'm not sure I'm a fan of those anyway, as they're not very machine readable, but I thought I'd mention it.

As part of the same effort we could also replace or supplement --verbose with a --log-level command-line option.

abingham:
I created #305 to follow through on this.

abingham:
> it precludes nice "graphical" progress displays in the terminal.

We might be able to accomplish these (if desired) with sufficiently advanced `Handlers` and `Formatters`. 

> replace or supplement --verbose with a --log-level command-line option.

I like this idea.

PR #315, opened 2017-11-26
UnittestRunner: remove useless-super-delegation

blueyed:
It was added in 96b20fb apparently in an attempt to please pylint it
seems.


abingham:
Thanks!

PR #316, opened 2017-11-26
setup.py: use PACKAGES list

blueyed:
This avoids calling `find_packages`.
It adds a test to ensure the list is what `find_packages` would return.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@ import io
 import os
 import sys
 
-from setuptools import setup, find_packages
+from setuptools import setup
 
 
 def local_file(*name):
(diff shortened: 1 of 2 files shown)


blueyed:
Not too sure about this altogether.

But having e.g. pytest-mock might be useful in the future in general.

abingham:
Yeah, I'm not sold on this change. What's the practical benefit?

blueyed:
Less overhead when `pip install -e .`ing.
But then more overhead when running the tests.

Closing for now.
How do you feel about pytest-mock?

abingham:
If have no problem with pytest-mock if it's something that we'll benefit from.

PR #319, opened 2017-11-26
Travis: remove install section

blueyed:
tox takes care of installation.

This removes the call to `cosmic-ray test-runners`, which should be
covered by a real integration test instead (if not done already).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,16 +11,7 @@ matrix:
   - python: nightly
 
 install:
-- python setup.py install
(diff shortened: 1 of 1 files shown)


abingham:
Ah, this is *very* nice. As I mentioned before, I'm still wrapping my head around tox, and you're right that we don't need to install the dependencies into the base system any more.

The call to `cosmic-ray test-runners` was some cruft from a earlier debugging effort, I think.

PR #320, opened 2017-11-26
Add logging to load_config

blueyed:
I've tried to run a command from the report ("cosmic-ray worker
module zero_iteration_loop 9") and it was waiting for input on stdin.

This additional logging should help in this case (when run in verbose
mode).

### DIFF
--- a/cosmic_ray/config.py
+++ b/cosmic_ray/config.py
@@ -1,20 +1,25 @@
 """Configuration module."""
+import logging
 import sys
 
 import yaml
+
+LOG = logging.getLogger()
 
 
 def load_config(filename=None):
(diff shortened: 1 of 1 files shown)


abingham:
My only concern with this is that it means that workers could be printing something besides JSON to stdout. But I guess that's not a problem in "production" runs since the workers won't be running in verbose mode.

abingham:
Thanks!

blueyed:
> could be printing something besides JSON to stdout

logging goes to stderr.

Checked with `cosmic-ray --verbose worker --keep-stdout covimerage zero_iteration_loop 9 config.yml >/dev/null`.

PR #321, opened 2017-11-26
Travis: remove unnecessary call to pyenv

blueyed:


abingham:
I admit I don't know why that was there in the first place!

blueyed:
It can help when you want to use multiple Python versions with the same job.

PR #323, opened 2017-11-26
Added a simple example demonstrating how to use cosmic-ray to improve a test suite.

abingham:
This is a rebasing and simplification of a set of changes by @trbedwards.


abingham:
@trbedwards Can you look this over and let me know if it looks correct to you?

examples/bowling_game_score_calculator/config.yml
```diff
@@ -0,0 +1,13 @@
+# Run the adam tests with unittest
```
blueyed: Comment needs to get updated?!

blueyed: Should this (bowling_game_score_calculator) be in here (this PR) after all?

abingham: That comment is cruft; I've removed it. 

I *think* the `bowling_game_score_calculator` was part of the same documentation effort from the PyCon UK sprint. @trbedwards: was this intended to be in the same PR as `simple_math`? I may have messed up the rebasing or something.

PR #325, opened 2017-11-26
is_killed: consider TestOutcome.INCOMPETENT to be killed

blueyed:


blueyed:
> The diff coverage is 0%.

PR #329, opened 2017-11-27
No longer building configs during init.

abingham:
This turned out to be to cutesy during actual use. Now you need to use
`new-config` (or some other technique) to create configs.


cosmic_ray/config.py
```diff
@@ -13,14 +13,23 @@ def load_config(filename=None):
...
+        try:
+            return yaml.safe_load(sys.stdin)
+        except yaml.parser.ParserError as exc:
+            raise ConfigError() from exc
 
-    with open(filename, mode='rt') as handle:
-        LOG.info('Reading config from %r', filename)
-        return yaml.safe_load(handle)
+    try:
+        with open(filename, mode='rt') as handle:
```
rob-smallshire: This is a text file, so there's an implicit default encoding here. The YAML spec (http://www.yaml.org/spec/1.2/spec.html#id2771184) implies we should use UTF-8 unless we want to support UTF-16 too through explicit byte order marks (I guess not).

test/unittests/test_config.py
```diff
@@ -0,0 +1,46 @@
...
+    temp_stdin.seek(0)
+    mocker.patch('sys.stdin', temp_stdin)
+
+    with pytest.raises(ConfigError):
+        load_config()
+
+
+def test_load_from_valid_config_file(tmpdir):
+    config_path = tmpdir / 'config.yml'
+    with config_path.open(mode='wt') as handle:
```
rob-smallshire: Here we should also set encoding='utf-8' when opening the file.

test/unittests/test_config.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def test_load_non_existent_file_raises_ConfigError():
+    with pytest.raises(ConfigError):
+        load_config('/foo/bar/this/does/no-exist/I/hope')
+
+
+def test_load_from_invalid_config_file_raises_ConfigError(tmpdir):
+    config_path = tmpdir / 'config.yml'
+    with config_path.open(mode='wt') as handle:
```
rob-smallshire: Here we should also set encoding='utf-8' when opening the file.

rob-smallshire: We could also have a test which writes some binary date which is not decodable as UTF-8 to ensure that the UnicodeDecodeError is handled correctly.

cosmic_ray/config.py
rob-smallshire: Apparently PyYAML screws up YAML with BOMs anyway, so we must assume UTF-8.

PR #331, opened 2017-11-27
Detect ast.Compare nodes where the right-hand-side value is None

rob-smallshire:
This change avoids creating a whole class of incompetent mutants by detecting when the right-hand side value of a comparison operator is None, and restricting the set of replacement mutant operators to only those which can operate on None.

### DIFF
--- a/cosmic_ray/operators/comparison_operator_replacement.py
+++ b/cosmic_ray/operators/comparison_operator_replacement.py
@@ -10,6 +10,7 @@ from ..util import build_mutations, compare_ast
 OPERATORS = (ast.Eq, ast.NotEq, ast.Lt, ast.LtE, ast.Gt, ast.GtE,
              ast.Is, ast.IsNot, ast.In, ast.NotIn)
 
+
 def _all_ops(from_op):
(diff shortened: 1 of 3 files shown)


abingham:
This looks good, esp. considering that we don't have a better alternative right now. In the anchoring/exception work I've been thinking of adding a phase (currently called "interception", though I don't remember where I got that name) after `init` which would filter out mutations that we didn't want to perform. We could be skipping them because of anchored metadata, but we could also use that phase to implement the kind of filtering you're doing here (though that raises the question of how to provide AST-level access during interception).

rob-smallshire:
I'll admit I don't grok this codecov report. The urgent-looking red rectangle pertains to `comparison_operator_replacement.py`, and this is also shown as having decreased coverage, yet according to the report I click through all the added lines are covered. One `for` statement is in yellow, but I don't know what that means. The statement inside the same look are green. Odd.

abingham:
I agree. I merged in the codecov stuff without giving it a lot of review, but I need to come to grips with what it's telling us.

PR #332, opened 2017-11-28
Added support for skipping mutations based on spor metadata.

abingham:
This is a first attempt at using [spor](https://github.com/abingham/spor)'s anchored metadata to disable mutations for specific lines of code. I think this particular commit is a bit of a minimum-viable-product insofar as a) it does provide crude control over mutation but b) could use quite a bit of work. 

Even in this unfinished state, I think it's valuable to merge into master for a few reasons. First, there are users who need this right now. Second, it gives us some experience with the approach. Finally, it gets rid of what had become a long-term feature branch, something I find cognitively unsettling and also burdensome from an engineering point of view. 

I plan to create some "request for comment"-style issues for discussing what to do next. These next steps include:
- Improving spor's support for updating anchors as code changes (this is more of a spor issue)
- Replacing the "interceptors" concept here with something more general. I'm imagining plugins that define new "subcommands" for CR which users can run when appropriate. For example, there might be a command `cosmic-ray tool spor <session>` which users run after `init` which updates the work-db, doing the work that "interceptors" do in this PR.
- Improving performance
- etc.


test/test_project/adam.py
```diff
@@ -127,3 +127,7 @@ def decorated_func():
         result = True
 
     return result
+
+
+if __name__ == '__main__':
+    x = 3
```
rob-smallshire: I don't understand the significance of this.

test/unittests/test_command_line_processing.py
```diff
@@ -148,7 +148,8 @@ def test_config_success_returns_EX_OK(lobotomize, local_unittest_config, session
 
 
 def test_dump_success_returns_EX_OK(lobotomize, local_unittest_config, session_file):
-    cosmic_ray.cli.main(['init', local_unittest_config, session_file])
+    errcode = cosmic_ray.cli.main(['init', local_unittest_config, session_file])
+    assert errcode == 0
```
rob-smallshire: Should this also be os.EX_OK, for consistency.

cosmic_ray/interceptors/spor.py
```diff
@@ -0,0 +1,32 @@
+"""An interceptor that uses spor metadata to determine when specific mutations
```
rob-smallshire: GitHub won't allow me to leave a review comment on the empty `interceptors/__init__.py` file, so I'll leave one here.

I don't yet have a clear idea of what an "interceptor" is. I'd appreciate a brief explanation and a good place for this could be the module docstring for the `interceptors` package.

test/test_project/adam.py
abingham: Ah right, I should probably comment it. The issue is that __main__ blocks are a canonical example of equivalent mutants: CR will happily mutate them, but they are effective untestable. So I added this block as something to "block" with spor; you'll notice that the anchors in `.spor` indicate these two lines.

test/unittests/test_command_line_processing.py
abingham: Definitely!

cosmic_ray/commands/init.py
```diff
@@ -3,11 +3,48 @@
...
 log = logging.getLogger()
 
 
+class WorkDBInitCore:
+    """Operator core that initializes a WorkDB for a specific module and operator.
+
+    The idea is to walk the AST looking for nodes that the operator can mutate.
+    As they're found, `visit_mutation_site` is called and this core adds new
+    WorkItems to the WorkDB. Use this core to populate a WorkDB by creating one
+    for each operartor-module pair and running it over the module's AST.
```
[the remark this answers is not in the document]
abingham: Ah, good catch :)

PR #334, opened 2017-11-28
Reworked test runner failure

abingham:
This makes TestRunnerFailure behave a little better with respect to producing strings. Now it uses Exception's __str__ and __repr__ to produce more meaningful strings.


abingham:
@blueyed Can you look this over real quick?

Issue #333, opened 2017-11-28
`NumberReplacer` should do both an increment and a decrement


abingham:
`NumberReplacer` currently just increments the number it finds. Based on feedback from @rob-smallshire, it should also do a decrement to better test boundary-condition tests.

abingham:
Fixed by #335.

PR #335, opened 2017-11-28
NumberReplacer now increments and decrements.

abingham:
Addresses #333.

### DIFF
--- a/cosmic_ray/operators/number_replacer.py
+++ b/cosmic_ray/operators/number_replacer.py
@@ -6,14 +6,25 @@ import ast
 from .operator import Operator
 
 
+# Maps mutation count to number offsets
+OFFSET_MAP = {
+    0: 1,
+    1: -1,
+}
+
+
 class NumberReplacer(Operator):
     """An operator that modifies numeric constants."""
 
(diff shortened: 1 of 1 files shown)


cosmic_ray/operators/number_replacer.py
```diff
@@ -6,14 +6,25 @@
 from .operator import Operator
 
 
+# Maps mutation count to number offsets
+OFFSET_MAP = {
+    0: 1,
```
rob-smallshire: Although syntactically redundant, I'm a sticker for using the unary plus when it aids comprehension, which it would for the increment value.

cosmic_ray/operators/number_replacer.py
```diff
@@ -6,14 +6,25 @@
 from .operator import Operator
 
 
+# Maps mutation count to number offsets
+OFFSET_MAP = {
```
rob-smallshire: Why can't this be a list, which is indexed by integers starting from zero anyway? If it's helpful to explicitly see the indexes, I'm happy with it being a dictionary.

abingham: Good idea! The dict is certainly overkill for this.

PR #340, opened 2017-11-29
tox: use --cov=.

blueyed:


blueyed:
Fails to report coverage (but the report from pytest-cov is there):

> No source for code: '/home/travis/build/sixty-north/cosmic-ray/test/test_project/adam'.
> Aborting report output, consider using -i.

PR #341, opened 2017-11-30
Avoid operator in rhs integer

rob-smallshire:
Avoid mutating comparison operators to the `in` or `not in` operators when the right-hand side operand is a literal integer, because `x in 5` will always fail with a `TypeError`, producing an incompetent mutant.

Avoid mutating comparison operators to the `is` or `not is` operators when the right-hand side operand is a literal integer, because `x == 5` will be equivalent to `x is 5`.

### DIFF
--- a/cosmic_ray/operators/comparison_operator_replacement.py
+++ b/cosmic_ray/operators/comparison_operator_replacement.py
@@ -55,6 +55,7 @@ def _rhs_is_none_ops(from_op):
             yield from value
             return
 
+
 def _rhs_is_integer_ops(from_op):
     assert isinstance(from_op, ast.AST)
     yield from _RHS_IS_INTEGER_OPS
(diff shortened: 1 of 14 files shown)


cosmic_ray/operators/comparison_operator_replacement.py
```diff
@@ -46,19 +46,33 @@ def _all_ops(from_op):
...
 def _rhs_is_none_ops(from_op):
     for key, value in _RHS_IS_NONE_OPS.items():
         if isinstance(from_op, key):
             yield from value
             return
 
 
+def _rhs_is_integer_ops(from_op):
+    assert isinstance(from_op, ast.AST)
+    yield from _RHS_IS_INTEGER_OPS
```
abingham: Couldn't this be a simple `return`?

PR #344, opened 2017-12-02
add instructions on how to add a new test runner

drorasaf:


abingham:
Your documentation is really good, thanks for putting it together. The only issue I have with it is that, rather than adding new test runner into the `plugins/test-runners` directory, users should more generally create them as separate projects. We might merge very common ones into the CR repository, but it's not a requirement; they can provide runners (or any of our plugins) via separate repositories. I think it's a good idea to use the existing test-runner plugins as examples, though.

docs/testrunners.rst
```diff
@@ -0,0 +1,182 @@
...
+The results are returned as a (success, result)
+tuple. `success` is a boolean indicating if the tests
+passed. `result` is any object that is appropriate to provide
+more information about the success/failure of the tests.
+
+Test Runners are provided as plugins
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Cosmic Ray is designed to be extended with arbitrary test runners provided
+by users. It dynamically discovers tets runners at runtime using the
```
abingham: 'test' instead of 'tets'.

docs/testrunners.rst
```diff
@@ -0,0 +1,182 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The initial layout for our package is like this:
+
+::
+
+    setup.py
+    cosmic_ray_another_runner/
+      __init__.py
+      version.py
```
abingham: It looks like `version.py` isn't used below, so let's remove it. I think it's good to leave `version.py` out since it's not necessary and might confuse some people.

drorasaf:
Removed notorious docs:)

PR #346, opened 2017-12-04
Make `build_mutations` amenable to dynamic class generation

abingham:
Soon we'll likely need to generate classes dynamically based on (language) operator pairings. For example, rather than have a single class that handles all binary operator mutation, we'll have one per pairing. I expect to use `build_mutations` as part of this work, so this change is about making `build_mutations` ready for it.

Before, `build_mutations` took a sequence of AST node instances and a sequence of types. This won't be possible when we're generating classes because we won't be doing an AST traversal, i.e. we won't have node instances.

### DIFF
--- a/cosmic_ray/operators/comparison_operator_replacement.py
+++ b/cosmic_ray/operators/comparison_operator_replacement.py
@@ -77,7 +77,7 @@ def _build_mutations(node):
         ops = _rhs_is_none_ops
     else:
         ops = _all_ops
-    return build_mutations(node.ops, ops)
+    return build_mutations(map(type, node.ops), ops)
 
 
 class MutateComparisonOperator(Operator):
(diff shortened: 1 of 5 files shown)


abingham:
@rob-smallshire I think you looked at code in this area recently. Could you have a look over this change?

cosmic_ray/operators/comparison_operator_replacement.py
```diff
@@ -48,7 +48,7 @@ def _all_ops(from_op):
 
 def _rhs_is_none_ops(from_op):
     for key, value in _RHS_IS_NONE_OPS.items():
-        if isinstance(from_op, key):
+        if from_op is key:
```
rob-smallshire: Is there a case for using issubclass() in this and other locations? That would seem to be more equivalent to the original isinstance() call, or do we need an exact type check?

abingham: From what I can see, there are no inheritance relationships we'd want to capture here. I went with `is` since it's more precise and less likely to result in surprises.

PR #351, opened 2017-12-08
Replaced operator plugins with operator provider plugins.

abingham:
The basic

### DIFF
--- a/cosmic_ray/operators/operator.py
+++ b/cosmic_ray/operators/operator.py
@@ -1,5 +1,6 @@
 "Implementation of operator base class."
 
+from abc import ABCMeta, abstractmethod
 import ast
 
 
@@ -53,3 +54,23 @@ class Operator(ast.NodeTransformer):
         return '{}({})'.format(
             self.__class__.__name__,
             ', '.join(args))
+
+
+class ReplacementOperator(Operator, metaclass=ABCMeta):
+    """Base class for operators that have a "from-operator" and a "to-operator".
+
+    This is primarily a mechanism to tell linters that these properties exist.
+    Several of our operators build subclasses dynamically, and they primarily
+    just configure different to/from-ops.
+    """
+    @property
+    @abstractmethod
+    def from_op(self):
+        """The operator to mutate from."""
+        pass
+
+    @property
+    @abstractmethod
+    def to_op(self):
+        """The operator to mutate to."""
+        pass
--- a/cosmic_ray/operators/provider.py
+++ b/cosmic_ray/operators/provider.py
@@ -1,0 +1,44 @@
+"""Operator-provider plugin for all core cosmic ray operators.
+"""
+
+import itertools
+
+from . import (binary_operator_replacement,
+               boolean_replacer,
+               break_continue,
+               comparison_operator_replacement,
+               exception_replacer,
+               number_replacer,
+               remove_decorator,
+               unary_operator_replacement,
+               zero_iteration_loop)
+
+
+_OPERATORS = {
+    op.__name__: op
+    for op
+    in itertools.chain(
+        binary_operator_replacement.operators(),
+        comparison_operator_replacement.operators(),
+        unary_operator_replacement.operators(),
+        (boolean_replacer.AddNot,
+         boolean_replacer.ReplaceTrueFalse,
+         boolean_replacer.ReplaceAndWithOr,
+         boolean_replacer.ReplaceOrWithAnd,
+         break_continue.ReplaceBreakWithContinue,
+         break_continue.ReplaceContinueWithBreak,
+         exception_replacer.ExceptionReplacer,
+         number_replacer.NumberReplacer,
+         remove_decorator.RemoveDecorator,
+         zero_iteration_loop.ZeroIterationLoop))
+}
+
+
+class OperatorProvider:
+    """Provider for all of the core Cosmic Ray operators."""
+
+    def __iter__(self):
+        return iter(_OPERATORS)
+
+    def __getitem__(self, name):
+        return _OPERATORS[name]
--- a/cosmic_ray/operators/unary_operator_replacement.py
+++ b/cosmic_ray/operators/unary_operator_replacement.py
@@ -3,19 +3,19 @@
 
 import ast
 
-from .operator import Operator
 from ..util import build_mutations
+from .operator import ReplacementOperator
+
 
 # None indicates we want to delete the operator
-OPERATORS = (ast.UAdd, ast.USub, ast.Invert, ast.Not, None)
+_AST_OPERATORS = (ast.UAdd, ast.USub, ast.Invert, ast.Not)
 
 
 def _to_ops(from_op):
-    """
-        The sequence of operators which `from_op` could be mutated to.
+    """The sequence of operators which `from_op` could be mutated to.
     """
 
-    for to_op in OPERATORS:
+    for to_op in _AST_OPERATORS + (None,):
         if to_op and from_op is ast.Not:
             # 'not' can only be removed but not replaced with
             # '+', '-' or '~' b/c that may lead to strange results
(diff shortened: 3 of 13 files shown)


abingham:
@rob-smallshire Sorry that this is kind of a beast.

rob-smallshire:
Can you recommend and order in which I should review these files? It will be much more efficient if I approach this from the right direction.

abingham:
Start with the changes in `provider.py`. This is the heart of the change. Then look at `plugins.py` to see how the providers are used.

In `operator.py` I added a new base class used by the various "operator replacement" mutation operators, so look that over next. The hard parts are `unary_`, `binary_`, and `comparison_replacement_operator. We now generate a class for each specific to-/from-operator pair, whereas before we had a single operator that did them all. The `comparison_replacement_operator` is probably the most difficult of the bunch since the operator's AST node is so complex.

Once you're past these, the rest can be done in any order, I think. They are things like tests, documentation, etc.

cosmic_ray/plugins.py
```diff
@@ -1,24 +1,48 @@
...
     Attrs:
-        name: The name of the plugin containing the operator class.
+        name: The name of the operator class.
 
-    Returns: The operator *class object* (i.e. not an instance) provided by the
-        plugin named `name`.
+    Returns: The operator *class object* (i.e. not an instance).
     """
-    return ExtensionManager('cosmic_ray.operators')[name].plugin
+    sep = name.index('/')
```
rob-smallshire: Can you use:

    provider_name, _, operator_name = name.partition('/')

here?

cosmic_ray/plugins.py
```diff
@@ -1,24 +1,48 @@
...
+    provider = OPERATOR_PROVIDERS[provider_name]
+    return provider[operator_name]
 
 
 def operator_names():
-    """Get an iterable of all operator plugin names."""
-    return ExtensionManager('cosmic_ray.operators').names()
+    """Get an iterable of all operator names."""
+    return (
+        '{}/{}'.format(provider_name, operator_name)
```
rob-smallshire: Possible alternative: '/'.join((provider_name, operator_name))

Nothing wrong with yours though.

cosmic_ray/operators/operator.py
```diff
@@ -53,3 +54,23 @@ def __repr__(self):
         return '{}({})'.format(
             self.__class__.__name__,
             ', '.join(args))
+
+
+class ReplacementOperator(Operator, metaclass=ABCMeta):
```
rob-smallshire: Consider:

    from abc import ABC

then just inherit from ABC, which is just an empty class with ABCMeta as its metaclass, so

    class ReplacementOperator(Operator, ABC):

cosmic_ray/operators/operator.py
```diff
@@ -53,3 +54,23 @@ def __repr__(self):
...
+
+    This is primarily a mechanism to tell linters that these properties exist.
+    Several of our operators build subclasses dynamically, and they primarily
+    just configure different to/from-ops.
+    """
+    @property
+    @abstractmethod
+    def from_op(self):
+        """The operator to mutate from."""
+        pass
```
rob-smallshire: Consider `raise NotImplementedError` unless you intend to return `None` in the case this implementation is called.?

cosmic_ray/operators/operator.py
```diff
@@ -53,3 +54,23 @@ def __repr__(self):
...
+
+    @property
+    @abstractmethod
+    def to_op(self):
+        """The operator to mutate to."""
+        pass
```
rob-smallshire: Consider `raise NotImplementedError` unless you intend to return `None` in the case this implementation is called?

cosmic_ray/operators/provider.py
```diff
@@ -0,0 +1,44 @@
+"""Operator-provider plugin for all core cosmic ray operators.
+"""
+
+import itertools
+
+from . import (binary_operator_replacement,
```
rob-smallshire: I'm developing a preference for explicit absolute imports, so `from cosmic_ray.operators import ...`. I think it's because it makes me think less.

It's only a preference though.

cosmic_ray/operators/unary_operator_replacement.py
```diff
@@ -3,19 +3,19 @@
...
 
 def _to_ops(from_op):
-    """
-        The sequence of operators which `from_op` could be mutated to.
+    """The sequence of operators which `from_op` could be mutated to.
     """
 
-    for to_op in OPERATORS:
+    for to_op in _AST_OPERATORS + (None,):
         if to_op and from_op is ast.Not:
```
rob-smallshire: If this means:

    if (to_op is not None) and (from_op is ast.Not):

I think it would be clearer with the explicit `is` operator and parentheses.  If this isn't what was intended, then either it's wrong or unclear, so needs changing anyway ;-)

cosmic_ray/operators/unary_operator_replacement.py
```diff
@@ -27,21 +27,51 @@ def _to_ops(from_op):
...
-        return self.visit_mutation_site(
-            node,
-            len(build_mutations([type(node.op)], _to_ops)))
+    def visit_UnaryOp(self, node):  # pylint: disable=invalid-name, missing-docstring
+        if isinstance(node.op, self.from_op):
+            return self.visit_mutation_site(node)
+        return node
 
-    def mutate(self, node, idx):
+    def mutate(self, node, _):
```
rob-smallshire: My IDE lights up like a Christmas tree when arguments are ignored with `_` if you're overriding a method. I work around this by asserting on the argument just inside the method. For example.

    assert idx >= 0

cosmic_ray/operators/unary_operator_replacement.py
```diff
@@ -27,21 +27,51 @@ def _to_ops(from_op):
...
+            return node.operand
+
+        node.op = self.to_op()
+        return node
+
+
+def _create_replace_unary_operator(from_op, to_op):
+    """Create a ReplaceUnaryOperator subclasses that mutates `from_op` to `to_op`.
+    """
+    return type(
```
rob-smallshire: This line will send most folks scurrying for the documention – including me. Would be much clearer if you extracted three variables `name`, `bases` and `dictionary` from this complex expression.

cosmic_ray/operators/unary_operator_replacement.py
```diff
@@ -27,21 +27,51 @@ def _to_ops(from_op):
...
+        'ReplaceUnaryOperator_{}_{}'.format(
```
rob-smallshire: I've seen something that looks like this format call several times during this review. Is there an `op_class_name` function waiting to be extracted?

cosmic_ray/operators/unary_operator_replacement.py
```diff
@@ -27,21 +27,51 @@ def _to_ops(from_op):
...
+
+
+# Create the mutation operator classes
+_MUTATION_OPERATORS = tuple(
+    _create_replace_unary_operator(_AST_OPERATORS[idx], to_op)
+    for idx, to_op in build_mutations(_AST_OPERATORS, _to_ops))
+
+
+# Add the mutation operator classes to the module namespace.
+for op in _MUTATION_OPERATORS:
```
rob-smallshire: Consider:

    globals().update(op.__name__, op for op in _MUTATION_OPERATORS)

cosmic_ray/plugins.py
```diff
@@ -1,24 +1,48 @@
 """Query and retrieve the various plugins in Cosmic Ray.
 """
 
+import logging
+
 from stevedore import driver, ExtensionManager
 
+log = logging.getLogger()
```
rob-smallshire: More general remark than just this change, but should be we using:

    logging.getLogger(__name__)

so we get hierarchically named loggers?

PR #358, opened 2018-03-03
Reduce ast / run on windows

franzhaas:
I had issues running cosmic-ray on windows and hypothesis with these changes (+ one in spor but i did not find upstream of spor so far)

The things i changed are.:
- replaced os.EX constants with hard numbers, as the constants are unix only
- remove annotations from ast, before i did that my tests failed with recursion depth exceeded

The second problem i had on linux with pypy3 5.10 as well.

### DIFF
--- a/cosmic_ray/cli.py
+++ b/cosmic_ray/cli.py
@@ -61,7 +61,7 @@ def handle_baseline(args):
         print(''.join(work_item.data))
         return 2
 
-    return os.EX_OK
+    return 0
 
 
 @dsc.command()
@@ -74,7 +74,7 @@ def handle_new_config(args):
     with open(args['<config-file>'], mode='wt') as handle:
         handle.write(config)
 
-    return os.EX_OK
+    return 0
 
 
 @dsc.command()
@@ -146,7 +146,7 @@ def handle_init(args):
             config,
             timeout)
 
-    return os.EX_OK
+    return 0
 
 
 @dsc.command()
@@ -160,7 +160,7 @@ def handle_config(args):
         config, _ = database.get_config()
         print(json.dumps(config))
 
-    return os.EX_OK
+    return 0
 
 
 @dsc.command()
@@ -176,7 +176,7 @@ def handle_exec(args):
         args.get('<session-file>'))
     cosmic_ray.commands.execute(session_file)
 
-    return os.EX_OK
+    return 0
 
 
 @dsc.command()
@@ -192,7 +192,7 @@ def handle_dump(args):
         for record in database.work_items:
             print(json.dumps(record))
 
-    return os.EX_OK
+    return 0
 
 
 @dsc.command()
@@ -221,7 +221,7 @@ def handle_counts(args):
           sum(itertools.chain(
               *(d.values() for d in counts.values()))))
 
-    return os.EX_OK
+    return 0
 
 
 @dsc.command()
@@ -233,7 +233,7 @@ def handle_test_runners(args):
     assert args
     print('\n'.join(cosmic_ray.plugins.test_runner_names()))
 
-    return os.EX_OK
+    return 0
 
 
 @dsc.command()
@@ -245,7 +245,7 @@ def handle_operators(args):
     assert args
     print('\n'.join(cosmic_ray.plugins.operator_names()))
 
-    return os.EX_OK
+    return 0
 
 
 @dsc.command()
@@ -286,7 +286,7 @@ def handle_worker(args):
 
     sys.stdout.write(json.dumps(work_item))
 
-    return os.EX_OK
+    return 0
 
 
 DOC_TEMPLATE = """{program}
@@ -338,16 +338,16 @@ def main(argv=None):
             exit_at_end=False)
     except docopt.DocoptExit as exc:
         print(exc, file=sys.stderr)
-        return os.EX_USAGE
+        return 64
     except FileNotFoundError as exc:
         print(exc, file=sys.stderr)
-        return os.EX_NOINPUT
+        return 66
     except PermissionError as exc:
         print(exc, file=sys.stderr)
-        return os.EX_NOPERM
+        return 77
     except ConfigError as exc:
         print(exc, file=sys.stderr)
-        return os.EX_CONFIG
+        return 78
 
 
 if __name__ == '__main__':
--- a/cosmic_ray/worker.py
+++ b/cosmic_ray/worker.py
@@ -21,6 +21,7 @@ from .mutating import MutatingCore
 from .parsing import get_ast
 from .testing.test_runner import TestOutcome
 from .work_item import WorkItem
+import ast
 
 log = logging.getLogger()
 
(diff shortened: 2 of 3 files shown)


abingham:
The error-code problem definitely needs to be addressed. As I mentioned on the spor discussion, windows just isn't on my radar most of the time, so this kind of thing slips through. I'm happy to take that part of the PR as-is.

The second part of the PR is still a bit too mysterious to me to accept it right away. Where was it entering a recursion death-spiral? I'd be comfortable accepting your patch if we could at least identify the real problem (i.e. I assume there's bug in CR or Python that's really to blame...we shouldn't need to strip type annotations). This would allow us to unblock you, and we could work to fix the underlying issue as we have time, ultimately removing the type-annotation stripping.

franzhaas:
The problem apears to be that the system recurses into __subclasscheck__, either from abc or typing module.

I have provided a minimal example that is able to reproduce the issue for me at.:
https://github.com/franzhaas/minimal_example_cosmic_ray_ast

I am, just like you, not convinced that this is a bug in cosmic ray, at all.

abingham:
Hmmm. Your example seems to work just fine - no recursion problem, 0% survival rate - on mac os. I'll boot into windows and give it a shot there.

abingham:
I've managed to run it on windows now, and everything seems to mostly work there. I get two survivors on your minimal project:
```
> cosmic-ray init config.yml deleteme.json
> cosmic-ray exec deleteme.json
> cosmic-ray dump deleteme.json | cr-report
job ID 5b7a36aaa88b4c0fbd860bd28f4bde06:survived:mini
command: cosmic-ray worker mini core/ReplaceComparisonOperator_NotEq_Gt 0
--- mutation diff ---
--- aC:\Users\austin\repos\minimal_example_cosmic_ray_ast\mini.py
+++ bC:\Users\austin\repos\minimal_example_cosmic_ray_ast\mini.py
@@ -4,7 +4,7 @@
 from functools import wraps

 def _check_value(item: int) -> int:
-    if ((item & 255) != item):
+    if ((item & 255) > item):
         raise ValueError('Trying to mem dump non byte data (0x00 - 0xFF)')
     return item


job ID 04fc941680f64fa4a1d3843e4631d905:survived:mini
command: cosmic-ray worker mini core/ReplaceComparisonOperator_NotEq_Lt 0
--- mutation diff ---
--- aC:\Users\austin\repos\minimal_example_cosmic_ray_ast\mini.py
+++ bC:\Users\austin\repos\minimal_example_cosmic_ray_ast\mini.py
@@ -4,7 +4,7 @@
 from functools import wraps

 def _check_value(item: int) -> int:
-    if ((item & 255) != item):
+    if ((item & 255) < item):
         raise ValueError('Trying to mem dump non byte data (0x00 - 0xFF)')
     return item


total jobs: 25
complete: 25 (100.00%)
survival rate: 8.00%
```

But I don't see a recursion depth problem. Do you think you could reduce the minimal problem even further, getting Cosmic Ray out of the way? If this really is a problem with Python's AST walking or something, we can probably find a small example that shows the issue. I'll be happy to help out if I can.

rob-smallshire:
Do we know which Python version is being used?

franzhaas:
Ok, that is odd... I tested so far on windows with python3.6.4, with the version of cosmic-ray based on the release v3.1.0, but the constants replaced by integer literals

The session.json in the example repository has the error information inside.

I also tried on python 3.6.4 on Linux, with both the released version and master, with the same result.

I was able to reduce the example further, but still using hypothesis and cosmic ray. Importing typing in the source code under test is enough to trigger the issue.

abingham:
Interesting. There are a few things we can look at. First, what is your system's recursion limit? This could potentially be set very low. I doubt this is the problem, but we should at least verify. You can find it by running `sys.getrecursionlimit()`.

Second, what happens if you make one of the offending mutations by hand and run your test suite, all outside of the control of CR? If we see the recursion issue there, then this could be a pytest issue.

The stack traces in your example don't seem to indicate a problem in CR, but I can't rule anything out yet. Thanks for working with us on this.

abingham:
I've got another question, and it's largely because I don't have any real experience with Python's new type annotations. My understanding is that, by default, Python doesn't use the annotations for anything; you have to explicitly apply tooling or something to do type checking. But in the stack traces in your `session.json` is appears that *something* related to type checking is going on, i.e. the `typing` module is where the recursion seems to go crazy.

Are you using some tool (e.g. a pytest or hypothesis plugin) to invoke the type checking? Maybe that's why we're seeing different results.

As I said, I'm pretty new to the type-checking support, so I apologize if I'm way off base.

abingham:
You're right that the type annotations are included in the AST that the mutation operators see, but it's not clear to me how that might cause what you're seeing. From what I can see, the mutation has been successfully applied and you're seeing a failure during pytest execution. That's why I'm puzzled about the presence of type-checking functions in the stack-trace; I assume they would only be there if type-checking was taking place.

I'll look into this a bit more. I've got some ideas about what might be going on, but it'll take some investigation.

abingham:
I'm now confused but confident. I managed to get into a state on my mac os system where I was seeing the same errors you were seeing. This let me dig into things a bit, and I saw that the recursion happened when I imported `typing`; I didn't even need to apply any type annotations, just import the module.

However, I destroyed my virtual environment and reconstructed it, and now the problem is gone. Hence my confusion. My guess at this point is that some interactions between pyenv, virtual environments, `__pycache__`s, any goofiness that `typing` introduces, and CR's own gynmastics, we can end up with a corrupt/incoherent environment.

@franzhaas Could you try creating a fresh virtual environment, removing any `__pycache__`s that might be getting picked up, and running the CR test again? I'm grasping at straws here a bit, but I'm glad that I was at least able to replicate the error.

franzhaas:
Ok, we are on the same page then. The tool is very usefull and "does what i needs to be donne", right now with this modification.

Thank you very much.

franzhaas:
Sry, your comment overtook my typing...

I tried to run it with a new pipenv, and i removed __pycache__ in the project directory, but i got the recursion issue again.

franzhaas:
I do believe i figured out what happens.

the typing.py code, does not like it to be executed twice. I copied the content of typing.py into a file, and.:

C:\Projects\cosmic-ray_minimal_example>py -m pipenv run ipython
Python 3.6.4 (v3.6.4:d48eceb, Dec 19 2017, 06:54:40) [MSC v.1900 64 bit (AMD64)]
Type 'copyright', 'credits' or 'license' for more information
IPython 6.2.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import douple_type

In [2]: from collections import Sequence

In [3]: isinstance(1, Sequence)
---------------------------------------------------------------------------
RecursionError

franzhaas:
I think i figured it out, and opened a new pull request with the now proposed solution,  #363

abingham:
Good work! I can replicate your result, so this certainly seems like the issue. I'm going to close this PR and shift the discussion over to #363. Thanks!

Issue #281, opened 2017-11-08
Mutation for multi-process systems


abingham:
I think we've got an issue with multi-process system like Cosmic Ray. Since we perform mutation in-process, only the process that we start ourselves will see the mutation. If the system under test spawns new processes that should be subject to the mutation (e.g. if the SUT spawns a copy of itself), the copy won't see the mutation.

What does this mean? We need to find a mutation mechanism that is extrinsic to any particular Python process. For example, for most cases we could modify code on the filesystem. Then any Python process in that environment would see the mutation. Unfortunately this doesn't work with modules that are provided as e.g. zip files.

abingham:
This should be addressed in some upcoming changes.

PR #363, opened 2018-03-06
Prevent typing double import

franzhaas:
This replaces "Reduce ast / run on windows #358"

The problem i experience is explainable by running the typing.py code twice. 

This is what i think happens.:
typing.py gets executed when the unmodified original test code gets loaded
mutation takes place
the preserve_modules context manager deletes typing from the module list 

the mutated code gets loaded, and runs typing again

From now on, on my system at least isinstance(x, Sequence) will trigger an infinit recursion, if there are not more problems. (Why this is not easily reproducible on mac, i do not know)

The proposed solution is to load typing, so it does not get deleted by preserve_modules.

### DIFF
--- a/cosmic_ray/worker.py
+++ b/cosmic_ray/worker.py
@@ -19,7 +19,7 @@ try:
                        # which we shall not do twice... by loading it here,
                        # preserve_modules does not delete it and therefore
                        # fancy stuff happens only once
-except:
+except ImportError:
     pass
 
 from .config import serialize_config
--- a/cosmic_ray/worker.py
+++ b/cosmic_ray/worker.py
@@ -14,6 +14,13 @@ import sys
 import traceback
 
 import astunparse
+try:
+    import typing      # the typing module does some fancy stuff at import time
+                       # which we shall not do twice... by loading it here,
+                       # preserve_modules does not delete it and therefore
+                       # fancy stuff happens only once
+except:
+    pass
 
 from .config import serialize_config
 from .importing import preserve_modules, using_ast


abingham:
This looks reasonable, at least as a short-term solution. It's small, and it seems unlikely to cause any problems with getting good testing results. 

In the long-term, I think this is another indicator that we need to change the way we perform mutation. #281 is another indicator. Basically, I think we need to perform mutation "on disk" rather than in process. This is not an easy problem, so until we have a solution I'm happy to accept this patch.

This PR needs a small amount of work; I'll note those in review comments.

cosmic_ray/cli.py
```diff
@@ -61,7 +61,7 @@ def handle_baseline(args):
         print(''.join(work_item.data))
         return 2
 
-    return os.EX_OK
+    return 0
```
abingham: Please use the new `cosmic_ray.exit_codes` module for the exit codes. See commit 56708ff1b6c9608070088327fd3e305d50284e0d for the details.

cosmic_ray/worker.py
```diff
@@ -14,6 +14,13 @@
 import traceback
 
 import astunparse
+try:
+    import typing      # the typing module does some fancy stuff at import time
+                       # which we shall not do twice... by loading it here,
+                       # preserve_modules does not delete it and therefore
+                       # fancy stuff happens only once
+except:
```
abingham: What is the exception that you expect here? The only exception I would expect here is an `ImportError` for Python 3.4 (the earliest version we support).

abingham:
Thanks!

franzhaas:
My pleasure,

This patch is indeed not good as a long term solution. For example this patch does not allow the usage of cosmic-ray for testing.py, or anything with fancy monkey patching.

Is there a chance that this improvement makes it into an official release soon? Maybe fast tracked as a bugfix release?

That would  make my life easier. (no pinning of versions...)

Thanks

franzhaas:
@abingham 

I do not want to anoy you, but you mentioned you tend to forget to make releases ocasionally, maybe this is the right time to send you a kind reminder?

abingham:
No worries. A bunch of work stuff landed at once, and I did lose track of
this (and even if I'd remembered, I had no free time!) I'll try to get one
out now.

abingham:
Ugh. I just tried to push up a new release to PyPI, but there are
apparently some technical issues on their end. I'll try to remember to do
it later, but don't hesitate to ping me again if I don't get it done today
or tomorrow.

abingham:
@franzhaas v3.1.2 is up on PyPI now. Let me know if you see any problems with it.

PR #376, opened 2018-05-25
Pytest xdist integration

franzhaas:
I had the follwing issue.:

testing test code that is used for testing code that uses cffi.memmove, i have many cases that trigger memory access faults.

That is not an issue with pytest, as using the xdist plugin and the -n1 parameter, memory access faults are just a new way to fail a test. 

I added pytest-xdist as install dependency and the -n1 flag for pytest in this branch, as it seems to me to  be a reasonable default.

A reasonable alternative would be to use  config options... however i do believe that would be nice to have built in

### DIFF
--- a/cosmic_ray/plugins.py
+++ b/cosmic_ray/plugins.py
@@ -41,8 +41,11 @@ def get_operator(name):
 
 
 def operator_names():
-    """Get an iterable of all operator names."""
-    return (
+    """Get all operator names.
+
+    Returns: A sequence of operator names.
+    """
+    return tuple(
         '{}/{}'.format(provider_name, operator_name)
         for provider_name, provider in OPERATOR_PROVIDERS.items()
         for operator_name in provider)
@@ -62,7 +65,10 @@ def get_test_runner(name, test_args):
 
 
 def test_runner_names():
-    """Get iterable of test-runner plugin names."""
+    """Get all test-runner plugin names.
+
+    Returns: A sequence of test-runner plugin names.
+    """
     return ExtensionManager(
         'cosmic_ray.test_runners',
         on_load_failure_callback=_log_extension_loading_failure,
@@ -84,7 +90,10 @@ def get_interceptor(name):
 
 
 def interceptor_names():
-    """Get an iterable of all interceptor plugin names."""
+    """Get all interceptor plugin names.
+
+    Returns: A sequence of interceptor plugin names.
+    """
     return ExtensionManager(
         'cosmic_ray.interceptors',
         on_load_failure_callback=_log_extension_loading_failure,
@@ -104,7 +113,10 @@ def get_execution_engine(name):
 
 
 def execution_engine_names():
-    """Get iterable of execution-enginer plugin names."""
+    """Get all execution-engine plugin names.
+
+    Returns: A sequence of execution-engine names.
+    """
     return ExtensionManager(
         'cosmic_ray.execution_engines',
         on_load_failure_callback=_log_extension_loading_failure,
--- a/plugins/test-runners/pytest/cosmic_ray_pytest_runner/runner.py
+++ b/plugins/test-runners/pytest/cosmic_ray_pytest_runner/runner.py
@@ -37,7 +37,10 @@ class PytestRunner(TestRunner):
         else:
             args = []
 
+        args.append('-n1')
+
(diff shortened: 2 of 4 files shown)


cosmic_ray/plugins.py
```diff
@@ -41,8 +41,11 @@ def get_operator(name):
 
 
 def operator_names():
-    """Get an iterable of all operator names."""
-    return (
+    """Get all operator names.
+
+    Returns: A sequence of operator names.
+    """
+    return tuple(
```
abingham: Is your PR based on something other than `master`? This change (and some of your others) looks to be identical to one we made on the `fix-init` branch.

plugins/test-runners/pytest/cosmic_ray_pytest_runner/runner.py
```diff
@@ -37,7 +37,10 @@ def _run(self):
         else:
             args = []
 
+        args.append('-n1')
+
         with StringIO() as stdout:
+            exit_code = None
```
abingham: Why do you need to initialize this?

plugins/test-runners/pytest/cosmic_ray_pytest_runner/runner.py
```diff
@@ -59,7 +62,7 @@ def _run(self):
                                 for r in collector.reports if r.failed])
             if exit_code == 2:
                 return (False, [(repr(r), r.longreprtext)
-                                for r in collector.reports if r.failed])
+                                    for r in collector.reports if r.failed])
```
abingham: This indentation seems unnecessary.

abingham:
I don't entirely understand the motivation for this change. Are you saying that by using pytest-xdist you can get some of your tests to finish properly? 

I don't know a whole lot about pytest-xdist, but are you sure that the mutations will actually be performed properly with the plugin? CR does not (currently) mutate code on disk, so if pytest-xdist is rsyncing or something then mutations won't be propagated.

franzhaas:
Hi abingham,

yes this is not based on master.
For me a solution using the config file (altough i failed at trying it) would be great too.

Like last time, this is the code that fixed my problem at hand, i hope we can come to a usefull result from here together.

My problem is essentially that i test code which can be simplified to:
----8<----8<----8<----8<----8<----8<----
import cffi

ffi = cffi.FFI()

def sut():
    myptr = ffi.new("char[100]")
    ffi.memmove(myptr,  ffi.from_buffer(bytes(100)), 50 + 50)
    return myptr
----8<----8<----8<----8<----8<----8<----

If a mutation changes 50 + 50  to 50 * 50, I get a memory access problem and the test may crash with  a segfault.

using pytest without cosmic ray the solution is.:

pytest -n1 

which makes pytest to run the tests in an isolated process. Crashing is a "new form" of failing and all is fine. 

I tried to apply this aproach to cosmic ray, and got happy with the result.

abingham:
When you use Cosmic Ray without your changes, what do you see? I assume that the worker subprocess dies, and I would expect the monitor process to see this and mark the mutant as incompetent.

abingham:
I was able to reproduce the segfaults you were creating, and I saw that CR misbehaved in their presence. Basically, CR was always expecting *some* kind of valid JSON output from the worker subprocesses. In the case of a segfault, of course, that wasn't happening. The JSONDecodeError that resulted wasn't able to be serialized into the workdb.

I've created the [dealing-with-segfaults](https://github.com/sixty-north/cosmic-ray/tree/dealing-with-segfaults) branch that should fix this. I'm not convinced that this is the best solution, but let me know if gets you moving without requiring pytest-xdist.

franzhaas:
I tried this branch, and it does work.  However the mutations which crash apear to be treated as killed.

This is what i want, as in my humble point of view a test that triggers a crash with a reasonable reliability is a valid failing test.

However It apears not to be what you expect.

I suggest to reject this pull request and to follow your aproach.

abingham:
> However the mutations which crash apear to be treated as killed...However It apears not to be what you expect.

Actually it sounds like we're in agreement. I fully expect CR to treat segfaulting tests as "killed". They should really probably be "incompetent", and I can work on that, but they certainly shouldn't be survivors.

If you're satisfied with approach on my branch, continue working from that for a while until I can merge it into master. I'll close the PR. Thanks!

PR #380, opened 2018-06-01
Dealing with segfaults

abingham:
In PR #376 @franzhaas found that we weren't dealing with the case where the worker subprocess segfaults. He found cases where a mutation could create a program that segfaults, in which case the worker doesn't produce any valid JSON output. We were prepared for that case (i.e. bad JSON), but we were handling it incorrectly. That's the main thing this PR fixes.


cosmic_ray/reporting.py
```diff
@@ -8,7 +8,7 @@
 def _print_item(work_item, full_report):
     data = work_item.data
     outcome = work_item.worker_outcome
-    if outcome in [WorkerOutcome.NORMAL, WorkerOutcome.EXCEPTION]:
+    if outcome in [WorkerOutcome.NORMAL, WorkerOutcome.ABNORMAL, WorkerOutcome.EXCEPTION]:
```
rob-smallshire: Minor: If you use a set here rather than a list, the Python compiler optimises the code.

cosmic_ray/reporting.py
```diff
@@ -17,7 +17,7 @@ def _print_item(work_item, full_report):
             work_item.module),
         'command: {}'.format(work_item.command_line or '')
     ]
-    if outcome == TestOutcome.KILLED and not full_report:
+    if outcome in (TestOutcome.KILLED, TestOutcome.INCOMPETENT) and not full_report:
```
rob-smallshire: Similarly, can use a literal set here

cosmic_ray/reporting.py
```diff
@@ -43,7 +45,7 @@ def is_killed(record):
     """
     if record.worker_outcome in {WorkerOutcome.TIMEOUT, WorkerOutcome.SKIPPED}:
         return True
-    elif record.worker_outcome == WorkerOutcome.NORMAL:
+    elif record.worker_outcome in (WorkerOutcome.ABNORMAL, WorkerOutcome.NORMAL):
```
rob-smallshire: Another opportunity for a literal set.

cosmic_ray/worker.py
```diff
@@ -28,18 +27,20 @@
 from .parsing import get_ast
 from .testing.test_runner import TestOutcome
 from .work_item import WorkItem, WorkItemJsonDecoder
+import cosmic_ray.compat.json
 
 log = logging.getLogger()
 
 
 class WorkerOutcome:
```
rob-smallshire: We could/should be using Enum here. It's part of the standard library from Python 3.4 onwards.

cosmic_ray/worker.py
```diff
@@ -28,18 +27,20 @@
 from .parsing import get_ast
 from .testing.test_runner import TestOutcome
 from .work_item import WorkItem, WorkItemJsonDecoder
+import cosmic_ray.compat.json
```
rob-smallshire: The mix of relative and absolute import here is a bit jarring. FWIW I have a strong preference for absolute imports. They just make me think less. Explicit is better that implicit, yada yada

abingham:
Thanks, @rob-smallshire! I'll work all of your notes into subsequent commits.

PR #442, opened 2019-05-22
HTML report

LLKoder:
Added the expanded/collapsed list in the HTML report.

### DIFF
--- a/src/cosmic_ray/tools/html.py
+++ b/src/cosmic_ray/tools/html.py
@@ -36,20 +36,20 @@ def _generate_html_report(db):
             doc.stag(
                 'meta',
                 name='viewport',
-                content='width=device-width, initial-scale=1, shrink-to-fit=no'
-            )
+                content='width=device-width, initial-scale=1, shrink-to-fit=no')
             doc.stag(
                 'link',
                 rel='stylesheet',
-                href='https://stackpath.bootstrapcdn.com/bootstrap/4.1.1/css/bootstrap.min.css',
-                integrity='sha384-WskhaSGFgHYWDcbwN70/dfYBj47jz9qbsMId/iRN3ewGhXQFZCSftd1LZCfmhktB',
+                href='https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css',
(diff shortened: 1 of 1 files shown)


abingham:
Overall I like the idea, esp. that I can quickly scroll through the report and see red and green bars without having to look at the output. However, I have a few issues that I'd like to address (or at least discuss):

1. When I click either the "Summary Info" or "Job List" buttons, the other section gets closed. That is, if I have "Job List" open and I click "Summary Info", the job list is closed. This was not what I was expecting, and I think others will find it surprising, too. Is there any way to not make these sections mutually exclusive? Or is there a good reason to keep it the way it is?

This same problems applies to individual jobs: if I open one, the currently open is closed. This is actually a much bigger problem because of issue 2.

2. I would really like an "expand all" option that opens up all of the jobs. This way I can quickly scan through all of the jobs and see if there is a systematic problem with CR. This kind of scanning is something I have to do quite a bit during CR development when, for example, I break something that causes all mutants to be killed accidentally.

LLKoder:
1. I thought about this question before, but I seemed that user saw this info already and didn't need to keep it at a screen, user looked it. So I decided to do so. In principle, you may be right, It useful in the jobs list. To compare jobs results. I'll try to redo this part.

2. It's a good idea, I skipped it in my mind. I'll try to make it.

Please tell me, will I have to add new PR after I'll update commit in my account? Or will current PR get the update?

abingham:
> Please tell me, will I have to add new PR after I'll update commit in my account? Or will current PR get the update?

If you commit to your repository and push to github, the changes should show up in this pull request. No need for a new PR. 

But you should do a little reading on the pull-request workflow on gihtub first...I might be mis-remembering exactly how it work.

LLKoder:
I hope I did everything right. I'm about the updated PR.

abingham:
It looks like you updated the PR correctly, but it looks like there's a bug in "expand all". This is what I get when I push that button:

Screenshot 2019-05-27 at 10 52 06

LLKoder:
Hm... I cannot see the issue.
What browser (version) and OS do you use?
Is JavaScript enabled?
Can you send me your sqlite file for tests?

LLKoder:
Maybe you use Firefox. I see that bug when I use Tor (Firefox).

LLKoder:
I have fixed the issue. Please try again.

abingham:
It works great now! Yes, I was looking at it on firefox, so that was probably the issue. Thanks for sorting it all out.

PR #457, opened 2019-09-25
Exclude modules

Alcolo47:
Implementation of exclude-module

I don't use here interceptor because I don't want to see skipped module in the list of cosmic-ray -v INFO init

### DIFF
--- a/src/cosmic_ray/modules.py
+++ b/src/cosmic_ray/modules.py
@@ -4,17 +4,25 @@ import glob
 from pathlib import Path
 
 
-def find_modules(module_path):
+def find_modules(module_path, excluded_modules):
     """Find all modules in the module (possibly package) represented by `module_path`.
 
     Args:
         module_path: A pathlib.Path to a Python package or module.
+        excluded_modules: List for glob of modules to exclude.
 
     Returns: An iterable of paths Python modules (i.e. *py files).
     """
-    if module_path.is_file():
+    if module_path.is_dir():
+        pyfiles = set(glob.glob('{}/**/*.py'.format(module_path), recursive=True))
+
+    else:
         if module_path.suffix == '.py':
-            yield module_path
-    elif module_path.is_dir():
-        pyfiles = glob.glob('{}/**/*.py'.format(module_path), recursive=True)
-        yield from (Path(pyfile) for pyfile in pyfiles)
+            pyfiles = {module_path}
+        else:
+            pyfiles = set()
+
+    excluded = set(f for excluded_module in excluded_modules
+                   for f in glob.glob(excluded_module, recursive=True))
+
+    yield from (Path(pyfile) for pyfile in pyfiles - excluded)
--- a/src/cosmic_ray/modules.py
+++ b/src/cosmic_ray/modules.py
@@ -4,7 +4,7 @@ import glob
 from pathlib import Path
 
 
-def find_modules(module_path, excluded_modules):
+def find_modules(module_path, excluded_modules=None):
     """Find all modules in the module (possibly package) represented by `module_path`.
 
     Args:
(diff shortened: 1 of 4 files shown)


abingham:
Would it be any more difficult to simply filter the return value of find_modules()? Currently find_modules() only has to do one thing: find modules. Adding a filtering mechanism to it seems unnecessary if we can use extrinsic methods to accomplish the same thing.

abingham:
On a related note: if we ultimately *do* decide to change the API for find_modules(), you'll need to update the tests.

tests/test_suite/unittests/test_find_modules.py
```diff
@@ -1,6 +1,7 @@
 from pathlib import Path
+from unittest import result
```
abingham: Why is this being imported?

tests/test_suite/unittests/test_find_modules.py
```diff
@@ -42,3 +43,28 @@ def test_finding_modules_py_dot_py_using_slashes_with_full_filename(data_dir):
...
     assert expected == results
+
+
+def test_small_directory_tree_with_excluding_files(data_dir):
+    paths = (('a', 'b.py'), ('a', 'py.py'),
+             ('a', 'c', 'd.py'))
+    excluded_modules = ['**/__init__.py']
+    expected = set(Path(*path) for path in paths)
+
+    from cosmic_ray.execution.local import excursion
```
abingham: Just import this once at the top of the file (and remove the import below as well).

tests/test_suite/unittests/test_find_modules.py
Alcolo47: good question ...

tests/test_suite/unittests/test_find_modules.py
Alcolo47: ok

src/cosmic_ray/modules.py
```diff
@@ -12,9 +12,27 @@ def find_modules(module_path):
 
     Returns: An iterable of paths Python modules (i.e. *py files).
     """
-    if module_path.is_file():
-        if module_path.suffix == '.py':
-            yield module_path
-    elif module_path.is_dir():
+    if module_path.is_dir():
```
abingham: Why have you restructured this function? It seems like you've just moved things around without changing what it does.

tests/test_suite/unittests/test_find_modules.py
```diff
@@ -1,6 +1,6 @@
 from pathlib import Path
-
-from cosmic_ray.modules import find_modules
+from cosmic_ray.modules import find_modules, filter_paths
+from cosmic_ray.execution.local import excursion
```
abingham: For tests, you should use the `path_utils` fixture to get `excursion`. Ultimate the excursion definitions in `path_utils` and `execution.local` should probably be combined.

src/cosmic_ray/modules.py
abingham: If this doesn't have any functional change, I'd like it changed back.

Alcolo47: Why two tests?
There is an 3rd cases (silently ignored) ? When I read this previous code, I wonder if symbolic is take into account (I had to read is_file() specification to know that is yes).

abingham: Path's can be things besides files and directories, e.g. fifos or sockets. The way it's coded, we silently ignore those possibilities, but I'm comfortable with that because a) the docs say what to pass in and b) the generator behaves in a well-defined way if we get any of those other types.

If anything, we should probably just add an `else` that raises a `ValueError` when we receive neither a file nor a directory.

abingham:
This all looks great, thanks!

Issue #83, opened 2015-04-18
Reconsider the relational-operator skip set


abingham:
In the relational-operator replacement operator code, we have a set of replacements which we explicitly prevent. We do this because it seems that the replacements in question do little other than create equivalent mutants.

However, I may be wrong in my reasoning. It may be that we need these replacements. Reconsider these when you have time to really sort it out.

----------------------------------------
- Bitbucket: https://bitbucket.org/austin_bingham/cosmic-ray/issue/22
- Originally reported by: [austin_bingham](http://bitbucket.org/austin_bingham)
- Originally created at: 2015-03-28T12:43:44.956

rob-smallshire:
I think we can detect many bothersome equivalent mutants at runtime - and in particular the set of mutants skipped in the current implementation. We can't do this statically, but we can do it dynamically during the init test run.

We could detect the default and common case that equality tests boil down to equality of identity (that is `a == b` is equivalent to `a is b`) as follows: We could replace the operator by a call to our own function. This function could check `type(a).__eq__ is object.__eq__` as the test is running, and if it is, discard this particular mutation as being equivalent.  We've need to take care to preserve the semantics of the operators, but this is doable.

PR #458, opened 2019-09-25
Exclude operators

Alcolo47:
Adding exclude-operators options because there is irrelevant operators ( < -> !=  is -> ==)

### DIFF
--- a/src/cosmic_ray/commands/init.py
+++ b/src/cosmic_ray/commands/init.py
@@ -1,10 +1,12 @@
 "Implementation of the 'init' command."
 import logging
 import uuid
+from collections import defaultdict
 
 from cosmic_ray.ast import get_ast, Visitor
 import cosmic_ray.modules
-from cosmic_ray.plugins import get_interceptor, interceptor_names, get_operator
+from cosmic_ray.plugins import get_interceptor, interceptor_names, \
+    get_operator, load_operators
 from cosmic_ray.work_item import WorkItem
 
 log = logging.getLogger()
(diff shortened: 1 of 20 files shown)


abingham:
Thanks for your work on this. This kind of feature is certainly a useful one, and in fact we've had [some discussion along these lines already](https://github.com/sixty-north/cosmic-ray/issues/83). Also see the the skipping we implement in `comparison_operator_replacement.py`.

However, I think we need to take a different approach to implementing it. I would like to see this implemented as an *interceptor* rather than baked in to the core operator handling code. I know the interceptor feature isn't very well documented, and I apologize for that - my time for CR is very limited right now. But I do think it's the right way to filter/skip operators.

Here's how I see it working:
1. Create an interceptor plugin that is responsible for finding mutations that should be skipped (i.e. using regular expressions as you've done or via some sort of configuration)
2. When it runs, it can scan the work-db for jobs that should not run and mark them as "skipped".
3. Its configuration will be via the normal configuration file (which it can access from the work-db during interceptor execution time).
4. Users can configure the interceptor and en/disable it using the standard configuration system.

You can look at `cosmic_ray.interceptors.spor` for an example of how this could work. 

I think you'll be able to re-use the core of what you've already written, but the interceptor-based implementation won't require touching nearly as many files. It will be nicely encapsulated inside the existing system.

If I have some time this weekend I'll see if I can sketch this out more completely for you. Let me know if you have any questions. And again, thanks for your work on this...hopefully I'll have some time soon to give CR more attention!

abingham:
> If I have some time this weekend I'll see if I can sketch this out more completely for you.

Based on your other PRs, I can see you're way ahead of me on understanding interceptors!

src/cosmic_ray/commands/init.py
```diff
@@ -56,7 +57,27 @@ def init(module_paths, work_db, config):
       work_db: A `WorkDB` instance into which the work orders will be saved.
       config: The configuration for the new session.
     """
-    operator_names = cosmic_ray.plugins.operator_names()
+
+    operator_names = list(cosmic_ray.plugins.operator_names())
+
+    if log.isEnabledFor(logging.INFO):
```
abingham: Why do we need this here? Isn't this what the "operators" command is for?

abingham: With the exception of this logging, I think this PR looks good.

Alcolo47: I don't understand the problem:
- Just this test? It is useless to build complex log trace is there is no log watchers.
- The entire log block: It's just cosmetic, but to see and understand all available operators, some classification seems useful.

abingham: My point is that we've already got the `cosmic-ray operators` command which does this. If you want to update the output format of `operators`, I'm open to that discussion. But this logging just seems to duplicate an existing feature unnecessarily.

Alcolo47: Ok , I have forgotten `cosmic-ray operators`. I will rollback this log

abingham:
This is great, thanks!

abingham:
I've merge this in as is, but I'd really like it if you could add a few tests. We just need some way to know if this functionality works and to detect regressions.

PR #460, opened 2019-09-25
Adding interceptor to disable mutation on line with # pragma: no mutate

Alcolo47:
According to developers point of views, we can choose spor and/or pragma_no_mutate

### DIFF
--- a/src/cosmic_ray/interceptors/pragma_no_mutate.py
+++ b/src/cosmic_ray/interceptors/pragma_no_mutate.py
@@ -1,0 +1,44 @@
+"""An interceptor that uses metadata: no mutate to determine when specific mutations
+should be skipped.
+"""
+import re
+from functools import lru_cache
+import logging
+
+from cosmic_ray.work_item import WorkerOutcome, WorkResult
+
+log = logging.getLogger()
+
+
+def intercept(work_db):
+    """Look for WorkItems in `work_db` that should not be mutated due to spor metadata.
+
+    For each WorkItem, find anchors for the item's file/line/columns. If an
+    anchor exists with metadata containing `{mutate: False}` then the WorkItem
+    is marked as SKIPPED.
+    """
+
+    @lru_cache()
+    def file_contents(file_path):
+        "A simple cache of file contents."
+        with file_path.open(mode="rt") as handle:
+            return handle.readlines()
+
+    re_is_mutate = re.compile(r'.*#.*pragma:.*no mutate.*')
+
+    for item in work_db.work_items:
+        lines = file_contents(item.module_path)
+        try:
+            line_number = item.end_pos[0] - 1
+            if item.end_pos[1] == 0:
+                line_number -= 1
+            line = lines[line_number]
+            if re_is_mutate.match(line):
+                work_db.set_result(item.job_id,
+                                   WorkResult(output=None,
+                                              test_outcome=None,
+                                              diff=None,
+                                              worker_outcome=WorkerOutcome.SKIPPED))
+        except Exception as ex:
+            raise Exception("module_path: %s, start_pos: %s, end_pos: %s, len(lines): %s" %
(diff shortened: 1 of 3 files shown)


boxed:
I suggest changing this to "no mutate" instead of "no mutation" as this is already what mutmut does. No need to make these incompatible!

Alcolo47:
It mistake in the title: The code filer # pragma: no mutate (as mutmut does)

boxed:
Oh. Good!

abingham:
This is really nice!

src/cosmic_ray/interceptors/pragma_no_mutate.py
```diff
@@ -0,0 +1,47 @@
...
+    is marked as SKIPPED.
+    """
+
+    @lru_cache()
+    def file_contents(file_path):
+        "A simple cache of file contents."
+        with file_path.open(mode="rt") as handle:
+            return handle.readlines()
+
+    re_is_mutate = re.compile(r'.*#.*pragma:.*no mutate.*')
```
abingham: What if this is inside a string literal?

src/cosmic_ray/interceptors/pragma_no_mutate.py
```diff
@@ -0,0 +1,47 @@
...
+from functools import lru_cache
+import logging
+
+from cosmic_ray.work_item import WorkerOutcome, WorkResult
+
+log = logging.getLogger()
+
+
+def intercept(work_db):
+    """Look for WorkItems in `work_db` that should not be mutated due to spor metadata.
```
abingham: Fix this docstring.

src/cosmic_ray/interceptors/pragma_no_mutate.py
```diff
@@ -0,0 +1,47 @@
...
+
+    re_is_mutate = re.compile(r'.*#.*pragma:.*no mutate.*')
+
+    for item in work_db.work_items:
+        if not item.module_path:
+            continue
+
+        lines = file_contents(item.module_path)
+        try:
+            line_number = item.end_pos[0] - 1
```
abingham: Don't we run the risk of `line_number` being negative? Won't that give the wrong results later on (i.e. on line 38)?

src/cosmic_ray/interceptors/pragma_no_mutate.py
Alcolo47: I know, I can have false positive. I can lost some mutation.
Have you a best algorihtm ?

src/cosmic_ray/interceptors/pragma_no_mutate.py
Alcolo47: Line numbers seem to be 1-based en column position 0-based.

src/cosmic_ray/interceptors/pragma_no_mutate.py
abingham: Is there a package you can use that parses the actual code so that it would understand python syntax? I guess you'd have to parse the entire module to do this correctly.

What you're doing might be state-of-the art...I'm honestly not sure. But I'm very reluctant to make unnecessary guesses.

abingham:
It seems that you merged a whole lot of other changes into this. I can't work with that...keep your PRs focused.

Alcolo47:
Oups. git branch == spaghetti monster. I'll reset/rebase and push -f

src/cosmic_ray/interceptors/pragma_no_mutate.py
abingham: What I'm going to do is accept your PR as-is, and then create a "research issue" to remind us to investigate alternative approaches to detecting this pragma. I think this PR is really good, so I'd like to get it in sooner rather than later.

abingham:
Thanks!

PR #462, opened 2019-09-27
Clone without virtualenv

Alcolo47:
Create working directory in /tmp without creating virtualenv.
Then we can run tests within current environment (environment where cosmic-ray was launched from).
For this, multiple Workspace classes was created and workspace instantiation is driven with a new parameter [cloning] workspace_type

### DIFF
--- a/src/cosmic_ray/config.py
+++ b/src/cosmic_ray/config.py
@@ -107,6 +107,11 @@ class ConfigDict(dict):
         "The 'cloning' section of the config."
         return self['cloning']
 
+    @property
+    def cloning_config_workspace_type(self):
+        "The 'workspace' section of the cloning section"
+        return self.cloning_config.get('workspace_type', 'cloned_with_virtualenv')
+
 
 @contextmanager
 def _config_stream(filename):
--- a/docs/source/cloning.rst
+++ b/docs/source/cloning.rst
@@ -16,6 +16,37 @@ You can configure cloning in your configuration TOML in the
 `cosmic-ray.cloning` section. At a minimum, you must have a
 `cosmic-ray.cloning.method` entry in your config.
 
+Possible entry into `[cosmic-ray.cloning]` are:
+
+- `workspace_type`
+- `method`
+- `command`
+
+workspace_type
+==============
+
+The `workspace_type` entry is the name of workspace method creation:
+Possible names are:
+
+- `cloned`: Workspace is built by cloning your sources with the `copy` method
+  (see bellow). Useful if your sources can be run in-place.
+- `cloned_with_virtualenv`: Workspace is built by cloning your sources like
+  previous method and create an local virtualenv environment locally.
+  Useful to deploy your sources.
+
+::
+
+ [cosmic-ray.cloning]
+ workspace_type = 'cloned'
+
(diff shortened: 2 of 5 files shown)


abingham:
> Then we can run tests within current environment (environment where cosmic-ray was launched from).

Why do we need to do this?

Alcolo47:
Because, for my project in don't need to deploy my sources with setup.py or anything else.
My working environment is fully operational to run tests. Then:
- I spent 2/3 minutes to reinstall all /tmp virtualenvs * (cpu core number ==4)
- This can fails because there is to much pip concurrency on pypi.org access (timeout can occurs)
This, to duplicate an already existing environment.
- And in my stack gitlab ci / dockerizer runner, I have a problem by mixing virtualenv and pipenv.

abingham:
Ok, that's reasonable justification for this feature. I would, however, like to simplify the implementation. The use of metaclasses is a bit over-engineered. I think we could replace it and the `get_workspace_class()` and `get_workspace()` classmethods with a much simpler module-scope function (e.g. `get_workspace()`) that returns an instance. 

As it is, the use of the metaclass means I have to think harder than necessary about the code to come to the conclusion that it's just a complicated way to dispatch between two alternative workspace classes.

Alcolo47:
Yes, today it's overkill.

I have in my brain a others type of workspaces:
- docker
- docker swarm
- Kubernete (but I not competent for this now)

I wrote anything about that yet, but a real factory mechnanism seems to me a good start.

Other factories possibilities are:
- With a register class decorator (Writing decorator is as readable as metaclasses).
- Using plugin system, but today, workspace api is too simple for all of those workspaces. Then giving the possibility to write a 3rd party's plugin is not desirable.

abingham:
> Using plugin system,

If there's really a need for many different kinds of workspaces, this is definitely my preference. It's how the rest of CR works, and it seems to work very well.

> today, workspace api is too simple for all of those workspaces

Right. In my experience the hard part of defining a new plugin type is figuring out what the proper abstraction/API is. It sounds like you've identified 4 or 5 examples, and that should be enough for us to start to design a reasonable plugin API for workspaces.

> Then giving the possibility to write a 3rd party's plugin is not desirable.

Maybe I misunderstand what you're saying, but I think letting 3rd parties write workspace plugins is very desirable.

Alcolo47:
> Maybe I misunderstand what you're saying, but I think letting 3rd parties write workspace plugins is very desirable.

It is very desirable is the base API is well defined. But it is not the case yet.

abingham:
> It is very desirable is the base API is well defined. But it is not the case yet.

Good, then we're in agreement :)

abingham:
Can I close this in favor of defining a plugin system for workspaces?

src/cosmic_ray/config.py
```diff
@@ -123,6 +123,11 @@ def badge_format(self):
     def badge_thresholds(self):
         return self.badge['thresholds']
 
+    @property
+    def cloning_config_workspace_type(self):
+        "The 'workspace' section of the cloning section"
```
mmphego, line 128:
```suggestion
         """The 'workspace' section of the cloning section."""
```

PR #464, opened 2019-09-30
Replace multiprocessing.ProcessPoolExecutor with asyncio

Alcolo47:
The main idea is this PR is to remove `pending = list(pending_work)` with asyncio version.
This allows starting some jobs without waiting for all work items fetch ware done from database.

### DIFF
--- a/src/cosmic_ray/execution/local.py
+++ b/src/cosmic_ray/execution/local.py
@@ -34,15 +34,19 @@ One major difference is the directory in which the subprocesses run. They run
 the root of the cloned repository, so you need to take this into account when
 creating the configuration.
 """
-
+import asyncio
+import concurrent.futures
 import contextlib
 import logging
 import multiprocessing
 import multiprocessing.util
 import os
+from typing import Iterable, Callable
 
 from cosmic_ray.cloning import ClonedWorkspace
+from cosmic_ray.config import ConfigDict
 from cosmic_ray.execution.execution_engine import ExecutionEngine
+from cosmic_ray.work_item import WorkItem
 from cosmic_ray.worker import worker
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


src/cosmic_ray/testing.py
```diff
@@ -74,6 +74,5 @@ def run_tests(command, timeout=None):
         asyncio.set_event_loop_policy(
             asyncio.WindowsProactorEventLoopPolicy())
 
-    result = asyncio.get_event_loop().run_until_complete(
-        _run_tests(command, timeout))
+    result = asyncio.run(_run_tests(command, timeout))
```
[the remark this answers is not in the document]
abingham: Just to follow up on this, the docs for `run` say:

> This function always creates a new event loop and closes it at the end. It should be used as a main entry point for asyncio programs, and should ideally only be called once.

If nothing else, you're calling it in two places in this PR. This feels like something we need to address.

src/cosmic_ray/execution/local.py
```diff
@@ -97,21 +101,23 @@ class LocalExecutionEngine(ExecutionEngine):
...
+                initargs=(config,)
+        ) as pool:
+
+            async def run_task(work_item):
+                result = await loop.run_in_executor(pool, _execute_work_item,
+                                                    work_item)
+                on_task_complete(*result)
+
+            tasks = [run_task(work_item) for work_item in pending_work]
+            await asyncio.gather(*tasks)
```
abingham: I'm a bit concerned about memory usage here. `tasks` is a list of the same size as the number of items in `pending_work`, potentially millions or even billions of items. When we destructure it on the call to `asyncio.gather()`, Python is going to create a complete copy of it into a tuple, essentially doubling the space we need for this function.

Is there some way we can tell asyncio to process everything lazily and report when they're all done?

src/cosmic_ray/execution/local.py
```diff
@@ -97,21 +101,23 @@ class LocalExecutionEngine(ExecutionEngine):
...
-                on_task_complete(job_id, result)
+                initargs=(config,)
+        ) as pool:
+
+            async def run_task(work_item):
+                result = await loop.run_in_executor(pool, _execute_work_item,
+                                                    work_item)
+                on_task_complete(*result)
+
+            tasks = [run_task(work_item) for work_item in pending_work]
```
abingham: Isn't this just `pool.map()`? The docs make some claims about `map()` achieving better performance by chunking the input; we should investigate that.

src/cosmic_ray/execution/local.py
```diff
@@ -97,21 +101,23 @@ class LocalExecutionEngine(ExecutionEngine):
     "The local-git execution engine."
 
     def __call__(self, pending_work, config, on_task_complete):
-        with multiprocessing.Pool(
+        asyncio.run(self._execute_pending_works(pending_work, config, on_task_complete))
```
abingham: See my other comments about `asyncio.run` below.

src/cosmic_ray/execution/local.py
Alcolo47: Yes I have replaced `pending` list against `tasks` list in memory. Where is the best ?

I think there are others problems:
- sqlite is not under asynio then pending_work generator may blocks all async jobs.
- Dead lock can occurs if we need to write result in db since pending_work db cursor is not ended.

abingham:
It sounds like we're trying to solve two problems at once now. The first problem is to start processing jobs before all work-items were fetched from the database. The second is to avoid needing to have large memory allocations (e.g. the `pending` and `tasks` lists). 

You mention an interesting point: sqlite is not using asyncio. Perhaps we can address both of these issues by using a database with an asyncio interface. Perhaps even aiosqlite. I would be very interested to see a solution like that, esp. if it can solve both of these problems.

PR #465, opened 2019-09-30
Don't mutate boolean operator inside type annotations

Alcolo47:
Don't mutate type annotation: ('or' on: `a: int or float = 1`)
For this, I introduce `ObjTest` to test easily many conditions in parso tree.

### DIFF
--- a/src/cosmic_ray/operators/boolean_replacer.py
+++ b/src/cosmic_ray/operators/boolean_replacer.py
@@ -1,7 +1,10 @@
 "Implementation of the boolean replacement operators."
 
 import parso.python.tree
+from parso.python.tree import Keyword, PythonNode, Operator as tree_Operator, \
+    Function
 
+from cosmic_ray.operators.util import ObjTest
 from .keyword_replacer import KeywordReplacementOperator
 from .operator import Operator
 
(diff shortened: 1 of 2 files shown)


abingham:
This is interesting. It makes me wonder: should we just prevent all mutation of type annotations? The annotations are only intended for external tools and should have no effect on runtime behavior. In that sense, mutations to type annotations will always be equivalent mutants: impossible to detect. Does that sound right?

Alcolo47:
Yes, I think we don't have to mutate annotation because it will always create a mutant.

Today, annotation is used for typing, and is used by tools to detect type mismatches.
But it is only convention, maybe some others use can be found.

To use them in application, annotations must be accessible by code, but without introspection, you can't. And to add information on variables, other ways already exist. I think in fact, that annotation handles and will handle only meta-information (like types) and never application information.

abingham:
Ok, it sounds like we agree that we can disable all mutation in type annotations. If that's the case, then we should abandon this PR and consider how to best do that. I think a new issue is where we should start; we can discuss the implementation strategy there. Does that sound ok to you?

Alcolo47:
If you have a look on the new PR pragma filter with interceptors, we can see my proposition of a new interceptor class.
With this new interceptor, we can easyly move annotation detection to a new interceptor. And we can enlarge this detection for other operators.

abingham:
Excellent, thanks, I'll take a look at that! 

> we can easyly move annotation detection to a new interceptor.

I could be convinced to put this detection into the core of CR. If we make a blanket policy of never mutating annotations, then that's an appropriate place for it. That way we never mutated annotations in the workdb; they would all be skipped all of the time, so they'd be noise.

It sounds like we can close the PR. If you agree, please do so.

Alcolo47:
I close this PR because I made new PR https://github.com/sixty-north/cosmic-ray/pull/475 based on class base interceptor.

PR #466, opened 2019-09-30
Operator remove named parameters

Alcolo47:
Remove named parameter from function call

Useful for django model filter for instance.

I added `# pragma: mutation-no-remove` to avoid deletion, because it can be useful to explicit
some parameters with theirs default values to allowing other mutation on its.

### DIFF
--- a/src/cosmic_ray/operators/binary_operator_replacement.py
+++ b/src/cosmic_ray/operators/binary_operator_replacement.py
@@ -54,12 +54,12 @@ def _create_replace_binary_operator(from_op, to_op):
 
 
 # Parent types of operators which indicate that the operator isn't binary.
-_NON_BINARY_PARENTS = set((
+_NON_BINARY_PARENTS = {
     'factor',       # unary operators, e.g. -1
     'argument',     # extended function definitions, e.g. def foo(*args)
     'star_expr',    # destructuring, e.g. a, *b = x
     'import_from',  # star import, e.g. from module import *
-))
+}
 
 
 def _is_binary_operator(node):
--- a/src/cosmic_ray/operators/boolean_replacer.py
+++ b/src/cosmic_ray/operators/boolean_replacer.py
@@ -15,7 +15,7 @@ class ReplaceTrueWithFalse(KeywordReplacementOperator):
 class ReplaceFalseWithTrue(KeywordReplacementOperator):
     """An that replaces False with True."""
     from_keyword = 'False'
-    to_keyword = 'To'
+    to_keyword = 'True'
 
 
 class ReplaceAndWithOr(KeywordReplacementOperator):
--- a/src/cosmic_ray/operators/provider.py
+++ b/src/cosmic_ray/operators/provider.py
@@ -6,7 +6,7 @@ import itertools
 from . import (binary_operator_replacement, boolean_replacer, break_continue,
                comparison_operator_replacement, exception_replacer,
                number_replacer, remove_decorator, unary_operator_replacement,
-               zero_iteration_for_loop)
+               zero_iteration_for_loop, remove_named_arguments)
 
 _OPERATORS = {
     op.__name__: op
@@ -21,7 +21,9 @@ _OPERATORS = {
         exception_replacer.ExceptionReplacer,
         number_replacer.NumberReplacer,
         remove_decorator.RemoveDecorator,
-        zero_iteration_for_loop.ZeroIterationForLoop))
+        zero_iteration_for_loop.ZeroIterationForLoop,
+        remove_named_arguments.RemoveNamedArguments,
+    ))
 }
 
 
--- a/src/cosmic_ray/operators/remove_named_arguments.py
+++ b/src/cosmic_ray/operators/remove_named_arguments.py
@@ -1,0 +1,46 @@
+"""Remove named arguments
+Remove in function call, named arguments
+
+Because in a lot of vase, named arguments represents options or fields.
+
+The lost of an option or a field must imply application failure.
+
+Examples:
+    - subprocess.Popen call
+    - ORM (like django models)
+"""
+
+from parso.python.tree import PythonNode, Operator as tree_Operator
+from cosmic_ray.operators.operator import Operator
+
+
+class RemoveNamedArguments(Operator):
+    def mutation_positions(self, node):
+        if isinstance(node, PythonNode) and node.type == 'arglist':
+            for child in node.children:
+                if isinstance(child, PythonNode) and child.type == 'argument':
+                    yield child.start_pos, child.end_pos
+
+    def mutate(self, node, index):
+        assert isinstance(node, PythonNode)
+        count = 0
+        for i, child in enumerate(node.children):
+            if isinstance(child, PythonNode) and child.type == 'argument':
+                if count == index:
+                    try:
+                        next_child = node.children[i + 1]
+                        if isinstance(next_child, tree_Operator) and next_child.value == ',':
+                            del node.children[i + 1]
+                    except IndexError:
+                        pass
+                    del node.children[i]
+                    return node
+                count += 1
+
+    @classmethod
+    def examples(cls):
+        return (
+            ("f(1, b=2)", "f(1,)"),  # Ending coma is acceptable
+            ("f(1, b=2, c=3)", "f(1, c=3)"),
+            ("f(a=1, b=2)", "f( b=2)"),
+        )
--- a/src/cosmic_ray/operators/remove_named_arguments.py
+++ b/src/cosmic_ray/operators/remove_named_arguments.py
@@ -9,38 +9,60 @@ Examples:
     - subprocess.Popen call
     - ORM (like django models)
 """
-
-from parso.python.tree import PythonNode, Operator as tree_Operator
+import re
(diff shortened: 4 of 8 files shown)


boxed:
I think we should have a conversation about more granular whitelist patterns. We'd need to agree on some syntax and category names. "Remove" is a bit vague! I'd suggest "#pragma: no mutate: category, category 2" as the general syntax (where # ends the list of categories to have compatibility with "# pragma: no cover"). The category in this case seems like it should be named "drop kwarg".

abingham:
I agree that "mutation-no-remove" and to general. There are all sorts of things mutations can remove besides keyword argument, e.g. decorators, exceptions handlers, base classes, etc. I'm much more in favor of a more general syntax along the lines of what @boxed describes. (And of course it would be great if CR and mutmut could use the same syntax).

src/cosmic_ray/operators/binary_operator_replacement.py
```diff
@@ -54,12 +54,12 @@ def examples(cls):
 
 
 # Parent types of operators which indicate that the operator isn't binary.
-_NON_BINARY_PARENTS = set((
```
abingham: Please please please stop putting spurious, unrelated changes into your PRs. My time for processing these is very limited, and whenever you do this I have to spend some time figuring out *why* you've done it. When the answer is "for no reason related to the actual change", I get frustrated.

src/cosmic_ray/operators/boolean_replacer.py
```diff
@@ -15,7 +15,7 @@ class ReplaceTrueWithFalse(KeywordReplacementOperator):
 class ReplaceFalseWithTrue(KeywordReplacementOperator):
     """An that replaces False with True."""
     from_keyword = 'False'
-    to_keyword = 'To'
```
abingham: While this *is* a bug, it should not be in this PR. I'll fix this myself. Thanks for pointing it out.

src/cosmic_ray/operators/provider.py
```diff
@@ -6,7 +6,7 @@
 from . import (binary_operator_replacement, boolean_replacer, break_continue,
                comparison_operator_replacement, exception_replacer,
                number_replacer, remove_decorator, unary_operator_replacement,
-               zero_iteration_for_loop)
+               zero_iteration_for_loop, remove_named_arguments)
```
abingham: Keep your imports in alpha order.

src/cosmic_ray/operators/provider.py
```diff
@@ -21,7 +21,9 @@
         exception_replacer.ExceptionReplacer,
         number_replacer.NumberReplacer,
         remove_decorator.RemoveDecorator,
-        zero_iteration_for_loop.ZeroIterationForLoop))
+        zero_iteration_for_loop.ZeroIterationForLoop,
+        remove_named_arguments.RemoveNamedArguments,
```
abingham: use alpha order here.

src/cosmic_ray/operators/remove_named_arguments.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def mutation_positions(self, node):
+        if ObjTest(node).match(PythonNode, type='arglist'):
+            for child in node.children:
+                if ObjTest(child).match(PythonNode, type='argument'):
+                    if not self._do_skip_argument(child):
+                        yield child.start_pos, child.end_pos
+
+    @classmethod
+    def _do_skip_argument(cls, node):
```
abingham: As discussed in the conversation, I'd rather handle this using a more general, interceptor-based approach. Generally speaking, operators should no be concerned with whether they're "disabled"; that decision is handled at a different place.

src/cosmic_ray/operators/util.py
```diff
@@ -19,3 +19,134 @@ def dec(cls):
...
+                write(',\n')
+            write(f"{indent}]")
+        write(')')
+        if not indent:
+            write('\n')
+
+    do_dump(node)
+
+
+class ObjTest:
```
abingham: I think this `ObjTest` system is very interesting and potentially very useful for other operators. If for some reason we don't accept this PR, could you submit this separately? It would be a shame to lose it.

src/cosmic_ray/operators/util.py
```diff
@@ -19,3 +19,134 @@ def dec(cls):
...
+            return self
+        return self._clone(self.obj(*args, **kwargs))
+
+    def __getitem__(self, item) -> 'ObjTest':
+        if self.obj is None:
+            return self
+        return self._clone(self.obj[item])
+
+
+class ObjTestOptionnal(ObjTest):
```
abingham: Should this be "ObjTestOptional" with one "n".

src/cosmic_ray/operators/util.py
```diff
@@ -19,3 +19,134 @@ def dec(cls):
...
+    >>> bool(ObjTest(node))
+
+    """
+    def __init__(self, obj):
+        self.obj = obj
+
+    def _clone(self, obj) -> 'ObjTest':
+        return type(self)(obj)
+
+    def match(self, clazz=None, **kwargs) -> 'ObjTest':
```
abingham: All of the public API needs docstrings.

abingham: Use `cls` instead of `clazz`. This what we use in almost all of CR.

abingham:
At a high level, I'm concerned about this operator. A lot of the time it will result in mutants that die simply because they violate the signature of the called function (i.e. because the removed keyword argument has no default value), not because the a test suite detected the failure. As a result, it will result in a lot of time spent doing work that doesn't add any value.

Are there ways we could tighten the operator to make it more useful? For example, could it only remove keyword arguments with no default value? Similarly, in order to avoid equivalent mutants, could it avoid removing keyword arguments where the argument's value is the same as the the default value?

Alcolo47:
Yes, generalization of # pragma no mutate: cat1, cat2 where catX can be names of operators (or shorter name than each operator can gives) is a good idea:

Now, I know how to read comment in line of a parso node, this is better way than 're.find("# no mutate", line)'. But for this I need a node: from interceptor, it's impossible (as is).

And because interpretation of # pragma lines are more complex, than "if exists, filter", 
I think we can integrate the pragma filter engine in the core: in WorkDBInitVisitor, for each mutation_positions results, check for pragmas, and then, add in db a result skipped entry.

src/cosmic_ray/operators/util.py
Alcolo47: cls is generally classmethod 1st argument, and is the current class. Here it can be any class.

boxed:
@abingham in the general case it's not really possible to know statically what function is being called in python. But we can make educated guesses.... but that seems like a lot of complexity.

abingham:
> But for this I need a node: from interceptor, it's impossible (as is).

@Alcolo47 An interceptor has access to the path to the module being mutated, and thus is fully capable of creating a parso node tree. Or it could just use the source code without needing to create an AST. With either of these in hand, an interceptor could determine if a particular work-item has a relevant pragma.

With clever caching, I bet we could do this fairly cheaply.  

> WorkDBInitVisitor, for each mutation_positions results, check for pragmas, and then, add in db a result skipped entry.

As you can probably tell, I'm fairly resistant to this idea. I see it as an issue of separation of concerns, and I haven't seen any argument that convinces me that we need to change the current implementation to handle pragmas. As far as I can tell, interceptors have all the power we need to do arbitrarily complex filtering of work items.

abingham:
> But we can make educated guesses.... but that seems like a lot of complexity.

@boxed You're right, of course, and I'm not advocating that complexity (yet). I just worry that we're discussing an operator that, for many (most?) users, creates more trouble than value. 

At this point I'm happy to add the operator. We have good support for disabling operators, so users can disable it if they don't like it. And as @Alcolo47 says, it's clearly useful in some cases.

src/cosmic_ray/operators/util.py
abingham: I use `cls` in any context where I want to use `class`. If I can't use `cls`, my alternative is `klass` (which I think we use in one place in the CR). For the sake of consistency within this codebase, I'd really prefer to use `cls` here.

boxed:
Do you have support for having it off by default but then you can enable it? That seems like a good trade off for this.

abingham:
> Do you have support for having it off by default but then you can enable it? 

I need to think about that. My objection to that is that it's surprising: it would be the first example of an operator that turned off by default. And since I'm only running on conjecture right now (i.e. as to this operator's utility), I'd rather keep things unsurprising until we know more. 

There have been other ideas by @rob-smallshire to have different "cohorts" of operators: high-value ones that most users will want to run all the time, less valuable ones that should only be run if explicitly enabled, etc. Maybe when we have time to flesh out an implementation for this, we can decide where this operator should live.

abingham:
> we can make educated guesses.... but that seems like a lot of complexity.

I can't help but wondering how complex this would actually be. <hand-waving>An interceptor could fire up a jedi server and ask it to make the hard guesses.</hand-waving> It seems less insane the more I think about it. (That's probably how insanity works.)

Alcolo47:
I close this PR. A new PR based on class based interceptor, pragma interceptor is open: https://github.com/sixty-north/cosmic-ray/pull/477

PR #468, opened 2019-10-01
Adding ASTQuery

Alcolo47:
ObjTest alone

### DIFF
--- a/src/cosmic_ray/operators/util.py
+++ b/src/cosmic_ray/operators/util.py
@@ -19,3 +19,134 @@ def extend_name(suffix):
         return cls
 
     return dec
+
+
+def dump_node(node, stdout=None):
+    import sys
+    stdout = stdout or sys.stdout
+    write = stdout.write
+
+    def do_dump(node, indent=''):
+        write(f"{indent}{type(node).__name__}({node.type}")
+        value = getattr(node, 'value', None)
+        if value:
+            value = value.replace('\n', '\\n')
+            write(f", '{value}'")
+        children = getattr(node, 'children', None)
+        if children:
+            write(', [\n')
+            for child in children:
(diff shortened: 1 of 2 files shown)


boxed:
Is this trying to solve the same problem as the ASTPattern class in mutmut? https://github.com/boxed/mutmut/blob/master/tests/test_mutation.py#L29 i.e. matching complex patterns on the AST.

It would be interesting to write the patterns we want to match using both systems and comparing how they read if so.

abingham:
This looks mostly fine to me. The only change I'd like is to put the code into its own module. `ObjTest` is closely associated with the AST, so I propose this:

1. Make `ast.py` into a package. `ast/__init__.py` can expose all of the same stuff it does today.
2. Put the `ObjTest` implementation into `ast/obj_test.py`.

With that, I think everything would be good.

Alcolo47:
OK, I will move this class. But about its name, I was not inspired, maybe AstTester ?

abingham:
> maybe AstTester?

How about ASTQuery? It seems like it's primarily for querying for subsets of an AST.

src/cosmic_ray/operators/util.py
```diff
@@ -19,3 +19,134 @@ def dec(cls):
...
+            return self
+        return self._clone(self.obj(*args, **kwargs))
+
+    def __getitem__(self, item) -> 'ObjTest':
+        if self.obj is None:
+            return self
+        return self._clone(self.obj[item])
+
+
+class ObjTestOptionnal(ObjTest):
```
abingham: Do you mean `ObjTestOptional`? There's only on 'n' in 'optional'.

abingham:
If you can fix the breaking tests, then I can merge this in.

PR #469, opened 2019-10-01
Integration of pragma filtering during init visit of nodes

Alcolo47:
According to @boxed idea for pragma category,
This is an implementation of pragma filtering of any operators at core level (not interceptors).

This implementation get comments directly from parso tree

### DIFF
--- a/src/cosmic_ray/ast.py
+++ b/src/cosmic_ray/ast.py
@@ -1,9 +1,12 @@
 "Tools for working with parso ASTs."
 
 from abc import ABC, abstractmethod
+from typing import Dict, List
 
 import parso.python.tree
 import parso.tree
+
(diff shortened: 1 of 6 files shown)


abingham:
Per [my comments elsewhere](https://github.com/sixty-north/cosmic-ray/pull/466#issuecomment-537034295), I'm not disposed to accept this approach without a fairly extraordinary justification.

Alcolo47:
This PR is just a prof of concept.

Ok in interceptor, you can if you call parso twice.

Another way is to enlarge interceptor concept to be callable during node visit or when a new work item is appended or can be appended (pre_hook, post_hook) to db.

Alcolo47:
See PR https://github.com/sixty-north/cosmic-ray/pull/474

PR #470, opened 2019-10-02
Read pragma annotation from source comment to filter jobs

Alcolo47:
Another Implementation with interceptor

Interceptors are now classes, a callback function is called on every new working_item found with its parso node.

### Important files to see:
- `interceptors/Interceptor.py`: Base Interceptor class
- Modification of `spor` (`intercept` is now `SporInterceptor.post_init`)
- `init.py`: Integrate interceptors call inside visits.
- The `pragma_no_mutate` interceptor is now obsolete with this

### Note:
This interceptor need to know for each operators, the name of its pragma category.
This name is computed by a property in the operator class (in generic way, based on class name). Like that, we can override this property to customize this name per operator.

I point out that operator class need to be changed to provide information needed by this interceptor.

### New interceptors can now:
- Avoid scan of modules (pre_scan_module_path can return False)
- Avoid insertion of worker_item
- Have post treatments after module visit. (Unused yet, but here for coherence)
- Have post treatments after insertion of worker_item: (Main goal of this change)
- As today, have post treatments after init (at end of visits).


src/cosmic_ray/interceptors/base.py
```diff
@@ -0,0 +1,31 @@
+from parso.tree import Node
+
+from cosmic_ray.operators.operator import Operator
+from cosmic_ray.work_db import WorkDB
+from cosmic_ray.work_item import WorkItem
+
+
+class Interceptor:
```
abingham: The only justification I can find for this class is that you can later avoid re-parsing of code in your PragmaInterceptor. How much time does that actually take? I suspect its negligible, and this is class structure you've created is unnecessary complexity. Can't the pragma work be done in the framework of the original interceptor API?

I'm not opposed to API changes, but a) I'd like some discussion of them beforehand and b) they needs justification.

src/cosmic_ray/ast.py
```diff
@@ -43,6 +47,40 @@ def get_ast(module_path, python_version):
...
+            # don't strip '\n'
+            prefix = node.prefix.strip(" \t")
+            if prefix:
+                return prefix
+
+        if isinstance(node, parso.python.tree.Newline):
+            return node.prefix
+
+
+def get_node_pragma_categories(node) -> None or Dict[str, None or List[str]]:
```
abingham: "pragma" is not a core CR concept. Any handling of pragmas needs to take place in the isolation of the appropriate interceptor.

src/cosmic_ray/commands/init.py
```diff
@@ -19,32 +25,39 @@ class WorkDBInitVisitor(Visitor):
     for each operator-module pair and running it over the module's AST.
     """
 
-    def __init__(self, module_path, op_name, work_db, operator):
-        self.operator = operator
+    def __init__(self, module_path, op_name, work_db, operator,
+                 interceptors: Interceptors):
```
abingham: This is not where interceptors belong.

src/cosmic_ray/operators/operator.py
```diff
@@ -13,12 +16,25 @@ class Operator(ABC):
...
         self._python_version = python_version
+        self._pragma_category_name = None
 
     @property
     def python_version(self):
         "Python major.minor version as a string."
         return self._python_version
 
+    @property
+    def pragma_category_name(self):
```
abingham: Pragmas have no place in the core CR code.

src/cosmic_ray/operators/util.py
```diff
@@ -19,3 +19,138 @@ def dec(cls):
         return cls
 
     return dec
+
+
+def dump_node(node, stdout=None):
```
abingham: Isn't this part of some other PR?

src/cosmic_ray/pragma.py
```diff
@@ -0,0 +1,77 @@
+import re
+from typing import Dict, List
+
+_re_pragma = re.compile(r'([-A-Za-z0-9](?: (?! )|[-A-Za-z0-9])*)(:?)(,?)')
+
+
+def get_pragma_list(line: str) -> None or Dict[str, None or List[str]]:
```
abingham: Pragmas are not part of core CR. This belongs in the interceptor plugin.

src/cosmic_ray/operators/operator.py
Alcolo47: That is precisely my note remark: This interceptor needs to know operators pragma-name, and it can be nice to configure this name per operator.
We can embed this name computations inside the interceptor, but there is no-way to have more user-friendly names.

src/cosmic_ray/operators/util.py
Alcolo47: Yes, but since that other PR isn't merged, I need to checkout those part of code here from the other PR. git can merge this without problems.

src/cosmic_ray/interceptors/spor.py
```diff
@@ -29,7 +28,7 @@ def file_contents(file_path):
             with file_path.open(mode="rt") as handle:
                 return handle.readlines()
 
-        for item in self.work_db.work_items:
+        for item in self.work_db.pending_work_items:
```
abingham: This is incorrect. Interceptors are very simple: after init, they get to manipulate the workdb. They don't care about pending work items, only all work items. 

And in any event, you shouldn't need to modify the spor interceptor to implement your own.

src/cosmic_ray/work_item.py
```diff
@@ -84,6 +84,9 @@ def __eq__(self, rhs):
     def __neq__(self, rhs):
         return not self == rhs
 
+    def __repr__(self):
```
abingham: This is really nice, and something I've intended to do for a while. I'll go ahead and pull this (and WorkItem.__repr__) over to master now so we don't lose it.

abingham: Small note though: We still maintain python-3.5 compatibility, so we can't use f-strings. Our Travis build will catch this, but it's something to remember for the future.

src/cosmic_ray/interceptors/spor.py
Alcolo47: pending_work_item filters all work items with results. Then I filter work_item potentially already skipped by another interceptor.

Alcolo47: pending_work_item will filter already skipped item. If another interceptor has already skipped a work_item, it is useless to analyze it again.

src/cosmic_ray/interceptors/base.py
Alcolo47: With this, we can easier analyze the node of the work_item.
Even if we compute AST again, we have to find the corresponding node for a given work_item. For line-based interceptors, it can be easy but to find an expression or argument node, it's harder.

src/cosmic_ray/commands/init.py
Alcolo47: Can you explain?

Alcolo47:
I close this PR because PR https://github.com/sixty-north/cosmic-ray/pull/476 will replace this one:
Better git branching (factorization of class based interceptor in a separate PR)

Same code, but operators still unchanged. Here I made `getattr(operator, 'pragma_category_name', None)` to allow operator to name itself its pragma category.

src/cosmic_ray/interceptors/base.py
abingham: > to find an expression or argument node, it's harder.
What kind of interceptor do you have in mind here?

src/cosmic_ray/commands/init.py
abingham: The core `init` procedure doesn't need to know anything about interceptors. Interceptors are intentionally a separate, isolated concept. I value this architectural decision more than the potential benefits of the API you've come up with.

src/cosmic_ray/operators/operator.py
abingham: Surely the pragma interceptor can generate reasonable names from provider names coupled with the operator class names.

src/cosmic_ray/interceptors/base.py
abingham: Ok, I see that your annotation interceptor makes a good case for needing access a this level. I'll need to think about this a bit.

src/cosmic_ray/commands/init.py
Alcolo47: This architecture extends the possibility of interceptors to intercept.
Init only know Interceptor interface. We always need to know interface, before interceptor interface was `callable(work_db)`.

PR #473, opened 2019-10-07
NoOp : Add outputs

Alcolo47:
**Correcting no_op name: NoOp**

Create baseline db with tempfile.mkstemp
cosmic_ray.commands.execute now needs user open db instead of filenaùe
On errors, Write errors in stdout and exit with exit code 1

### DIFF
--- a/src/cosmic_ray/cli.py
+++ b/src/cosmic_ray/cli.py
@@ -157,6 +157,18 @@ def handle_baseline(args):
 
     Runs a baseline execution that executes the test suite over
     unmutated code.
+
+    options:
+      --force      Force write over baseline session file
+                   if this file was already created by a previous
+                   run.
+      --report     Print the report result of this baseline run.
+                   If the job has failed, jobs's outputs will be
+                   displayed.
+
+    return code:
+        0 if the job has exited normally, else 1.
+
     """
     session_file = Path(args.get('<session-file>'))
     force = args.get('--force', False)
--- a/src/cosmic_ray/cli.py
+++ b/src/cosmic_ray/cli.py
@@ -28,7 +28,7 @@ from cosmic_ray.progress import report_progress
 from cosmic_ray.version import __version__
 from cosmic_ray.work_db import WorkDB, use_db
 from cosmic_ray.work_item import WorkItem, WorkItemJsonEncoder, WorkResult, \
-    WorkerOutcome, TestOutcome
+    TestOutcome
 
 log = logging.getLogger()
 
@@ -148,18 +148,22 @@ def handle_exec(args):
     session_file = args.get('<session-file>')
     with use_db(session_file, mode=WorkDB.Mode.open) as work_db:
         cosmic_ray.commands.execute(work_db)
-
     return ExitCode.OK
 
 
 @dsc.command()
 def handle_baseline(args):
-    """usage: cosmic-ray baseline <session-file>
+    """usage: cosmic-ray baseline [--force] [--report] <session-file>
 
     Runs a baseline execution that executes the test suite over
     unmutated code.
     """
     session_file = Path(args.get('<session-file>'))
+    force = args.get('--force', False)
+    dump_report = args.get('--report', False)
+
+    baseline_session_file = session_file.parent / '{}.baseline{}'.format(
+        session_file.stem, session_file.suffix)
 
     # Find arbitrary work-item in input session that we can copy.
     with use_db(session_file) as db:  # type: WorkDB
@@ -171,9 +175,15 @@ def handle_baseline(args):
 
         config = db.get_config()
 
+    if force:
+        try:
+            os.unlink(baseline_session_file)
+        except OSError:
+            pass
+
     # Copy input work-item, but tell it to use the no-op operator. Create a new
     # session containing only this work-item and execute this new session.
-    with use_db(mode=WorkDB.Mode.create) as db:
+    with use_db(baseline_session_file, mode=WorkDB.Mode.create) as db:
         db.set_config(config)
 
         db.add_work_item(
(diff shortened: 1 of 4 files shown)


Alcolo47:
With this, I can test by mutation environment, but:
- (cpu core number) workspace are initialized, to use only one, only once.

Note: I make baseline db in /tmp because if you launch baseline twice, we reuse the same db: key violation error. Then we have to delete it after the run.

src/cosmic_ray/cli.py
```diff
@@ -173,7 +173,7 @@ def handle_baseline(args):
 
     # Copy input work-item, but tell it to use the no-op operator. Create a new
     # session containing only this work-item and execute this new session.
-    with use_db(baseline_session_file, mode=WorkDB.Mode.create) as db:
+    with use_db(mode=WorkDB.Mode.create) as db:
```
abingham: I would prefer an option to keep the baseline database around for post-mortem analysis. In fact, I'd like that to be the default behavior, but we could have a flag to `baseline` that says to delete the database on exit.

src/cosmic_ray/cli.py
```diff
@@ -185,10 +185,18 @@ def handle_baseline(args):
...
+        cosmic_ray.commands.execute(db)
+
+        result: WorkResult = next(db.results)[1]
+        if result.test_outcome == TestOutcome.KILLED:
+            print("Execution with no mutation gives those following errors:")
+            for line in result.output.split('\n'):
+                print("  >>>", line)
+            return 1
+        else:
+            print("Execution with no mutation works fine:")
```
abingham: I try to follow the policy of not producing any output on success, so let's remove this print statement. (The printing on error is fine.)

src/cosmic_ray/commands/execute.py
```diff
@@ -30,30 +30,29 @@ def _report_progress(stream):
 
 
 @reports_progress(_report_progress)
-def execute(db_name):
+def execute(work_db):
     """Execute any pending work in the database stored in `db_name`,
```
abingham: Please update this docstring.

src/cosmic_ray/commands/execute.py
```diff
@@ -30,30 +30,29 @@ def _report_progress(stream):
...
+            log.info("Job %s complete", job_id)
+
+        log.info("Beginning execution")
+        engine(
+            work_db.pending_work_items,
+            config,
+            on_task_complete=on_task_complete)
+        log.info("Execution finished")
 
     except FileNotFoundError as exc:
```
abingham: Where is this `FileNotFoundError` going to come from? It used to come from `use_db`, but that's not an issue now.

src/cosmic_ray/operators/provider.py
```diff
@@ -37,7 +37,7 @@ def __iter__(self):
         return iter(_OPERATORS)
 
     def __getitem__(self, name):
-        if name == 'no_op':
+        if name == 'NoOp':
```
abingham: This seems like a separate change.

src/cosmic_ray/work_db.py
```diff
@@ -277,8 +278,16 @@ def use_db(path, mode=WorkDB.Mode.create):
...
+        path = tmp_path
+    else:
+        tmp_path = None
+
     database = WorkDB(path, mode)
     try:
         yield database
     finally:
         database.close()
+        if tmp_path:
```
abingham: This will not get called if `WorkDB.__init__()` throws an exception.

I think a better solution is to have a context manager that returns either a temp file or the `path` argument. It can use the TempFile context manager internally. This context manager can be wrapped around the entire body.

abingham:
I like the basic idea of your change to `execute`. It's a better API.

abingham:
> Note: I make baseline db in /tmp because if you launch baseline twice, we reuse the same db: key violation error. Then we have to delete it after the run.

Another option is to allow `baseline` (perhaps with a flag) to overwrite the existing baseline. As I mentioned in my comments, I think there's a benefit to keeping the baseline session around after the baseline command finishes, at least in some cases.

src/cosmic_ray/cli.py
Alcolo47: Reversed

src/cosmic_ray/commands/execute.py
Alcolo47: Removed. I don't re-raise anywhere `FileNotFoundError`, but I had changed message of the 1st.

src/cosmic_ray/cli.py
Alcolo47: print under --report

src/cosmic_ray/operators/provider.py
Alcolo47: It's just a bug correction: baseline claims for 'NoOp' and the naming convention for operator  names is CamelCase (coming from class name convention).

src/cosmic_ray/cli.py
```diff
@@ -185,10 +195,20 @@ def handle_baseline(args):
...
+        cosmic_ray.commands.execute(db)
 
-    return ExitCode.OK
+        result = next(db.results)[1]  # type: WorkResult
+        if result.test_outcome == TestOutcome.KILLED:
+            if dump_report:
+                print("Execution with no mutation gives those following errors:")
+                for line in result.output.split('\n'):
+                    print("  >>>", line)
+            return 1
```
abingham: You should use an appropriate enum from ExitCode here.

Alcolo47: `cosmic-ray baselines` means: "Does my program runs correctly inside cr system ?"
Return code convention wants that possible answers will be: 0 for yes or 1 for no.
All others return codes conventions are to tells "I can't answer correctly because ...".
Then, for me, there is no better returns that 1.

abingham: OK, I can live with this.

PR #474, opened 2019-10-08
Class based interceptors

Alcolo47:
Isolation of class based interceptor made in other PR.

### DIFF
--- a/docs/source/interceptors.rst
+++ b/docs/source/interceptors.rst
@@ -1,6 +1,8 @@
 Interceptors
 ============
 
+Usage
+-----
 In Cosmic Ray we use *Interceptors* to change the list of planified works.
 
 Interceptor can be activated in your config file with:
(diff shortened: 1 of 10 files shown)


abingham:
I've given this a lot of thought, and I'm not going accept this change to the interceptor API. I simply don't think it's warranted based on any actual need, and it damages the conceptual integrity of an otherwise well-running and simple system. 

Ignoring the "annotations" interceptor for a second, all of the other interceptors are easily implementable in the existing framework. I'm more than happy to consider PRs that do that (while taking into account other, specific issues I have with how they're implemented, e.g. that pragmas must be a self-contained change).

For annotations, this is a bigger topic, and one that may or may not involve interceptors. It requires a discussion before implementation to decide on a course of action. I've opened an issue to discuss it.

For now, I'm going to close the PRs that use the class-based interceptors.

PR #487, opened 2019-10-20
Speed up the init command

tomato42:
The `cosmic-ray init` command takes a lot of time, but that time is mostly spent on waiting for disk synchronisation, not generation of the actual mutation test cases.

this speeds up init execution from over 5 minutes to around 40s on my SSD (and makes it CPU bound, not disk bound)

the last optimisation is rather extreme ("don't use db synchronisation for initial case creation"), but as I've noted in the commit message, the process is not restartable anyway, so there's little point in strict synchronisation

### DIFF
--- a/src/cosmic_ray/commands/init.py
+++ b/src/cosmic_ray/commands/init.py
@@ -65,6 +65,7 @@ def init(module_paths, work_db, config):
     work_db.set_config(config=config)
 
     work_db.clear()
+    work_db.disable_synchronisation()
 
     for module_path in module_paths:
         module_ast = get_ast(
@@ -78,6 +79,7 @@ def init(module_paths, work_db, config):
 
     enabled_interceptors = config.sub('interceptors').get('enabled', ())
     apply_interceptors(work_db, enabled_interceptors, config)
+    work_db.enable_synchronisation()
 
 
 def apply_interceptors(work_db, enabled_interceptors, config):
--- a/src/cosmic_ray/commands/init.py
+++ b/src/cosmic_ray/commands/init.py
@@ -27,21 +27,24 @@ class WorkDBInitVisitor(Visitor):
         self.occurrence = 0
 
     def visit(self, node):
-        for start, stop in self.operator.mutation_positions(node):
-            self._record_work_item(start, stop)
+        self.work_db.add_work_items(
+            self._get_work_item(start, stop)
+            for start, stop in self.operator.mutation_positions(node))
+
         return node
 
-    def _record_work_item(self, start_pos, end_pos):
-        self.work_db.add_work_item(
-            WorkItem(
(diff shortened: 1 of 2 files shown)


boxed:
How many mutants is in your test scenario?

tomato42:
> How many mutants is in your test scenario?
11645

abingham:
The only part that concerns me here is that we have to expose the notion of "synchronization" in the workdb; it's an implementation detail leaking out. However, the performance improvements you're reporting are important...I'm honestly surprised that it takes that long for only 11K entries.

I'll accept this for now, but I'd like to consider a way that we could keep the notion of "synchronization" hidden in the workdb. This way we don't limit ourselves unnecessarily to changing implementations in the future. Perhaps we can arrange the init phase so that a single generator function produces all of the work items and they're all piped into the workdb as one sequence. The work-db could then manage synchronization, transactions, etc. entirely internally.

tomato42:
> all of the work items [cold be] piped into the workdb as one sequence. 

that's what I was initially thinking of doing, but then the sqlite buffer needs to be managed (to be big enough) and requires quite a bit shuffling around in the api to produce sequences instead of sequence of calls...
and the calls can be easily turned into no-ops if the backend db changes and doesn't require it any more

abingham:
Right, I agree that this isn't a critical issue, just something that I'd like to reconsider when I've got some breathing space. Actually, the idea of restructuring the init logic to produce a single sequence of WorkItems is appealing...it decouples the idea of "find all work items" from "put them in the workdb". Maybe we could alleviate the complexity at the workdb level (i.e. the buffer management, etc.) by writing the workitems down in chunks of an established size, say 100 or 1000 or whatever.

In any event, thanks for you work on this!

abingham:
@tomato42 I've was poking around in this area again, and I've got a change to init that seems to make things even faster, while at the same time addressing my concern about exposing the "sync" API on WorkDB. It's in the [faster-init](https://github.com/sixty-north/cosmic-ray/tree/faster-init) branch. If you have some time, could you try it out on your project?

tomato42:
@abingham 
with current master (2348de2ce9):
```
20.39user 0.14system 0:21.50elapsed 95%CPU (0avgtext+0avgdata 33580maxresident)k
3944inputs+19512outputs (6major+10511minor)pagefaults 0swaps
```
with faster-init (0b92f978769):
```
4.68user 0.02system 0:05.58elapsed 84%CPU (0avgtext+0avgdata 33624maxresident)k
0inputs+4816outputs (0major+10008minor)pagefaults 0swaps
```
so looks much better, but probably could be even faster, as it looks disk-bound again :)

Both of them generated 7714 jobs.

abingham:
Thanks for testing it out. I'll merge it in to master soon.

PR #491, opened 2019-11-02
Estimate mutation score by sampling

### LINKED ISSUE #490: Estimating mutation score by sampling
I was thinking about the issue #484, and I think there is a bit of a problem with it: what if the test coverage changes but the code does not? How to see if the changes to test coverage are ok, when no application code was changed?

Maybe the cosmic-ray could execute a random selection of the tests and calculate confidence interval for it?

The formula I found, uses:
_p_ for percentage of mutants that survived
_q_ for percentage of mutants that were killed (i.e. 1 - _p_)
_n_ number of mutants tested
_N_ number of mutants total
_z_ z-score (scaling factor for the given confidence level, 1.65 for 90%, 1.96 for 95%, 2.58 for 99%)

So if I have 8000 mutants, tested randomly 40 of them, 20% of them survived and I want to know a 95% confidence interval for that 20% I calculate:
sqrt((_p_ * _q_)/_n_) * _z_ * (1 - sqrt(_n_/_N_)) = sqrt(0.2 * 0.8 / 40) * 1.96 * (1 - sqrt(40/8000)) = 0.115

so by executing 40 tests, I know that the real mutation score of this test suite is 20% ± 11.5% (95% confidence)

The nice thing is that if the execution was selecting the tests at random, that estimation could be simply a switch to `cr-report` to base it off of total jobs vs complete jobs and already calculated survival rate.

tomato42:
Since a non-trivial application will have thousands of mutants, we can execute only some part of it (either based on time or count) and calculate the confidence interval for the estimated mutation score.

Change the runner to execute mutations in random order, extend `cr-rate` with ability to use the estimated mutation score and make it useful for use in CI systems.

fixes #490

### DIFF
--- a/src/cosmic_ray/tools/survival_rate.py
+++ b/src/cosmic_ray/tools/survival_rate.py
@@ -1,6 +1,7 @@
 "Tool for printing the survival rate in a session."
 
 import math
+import sys
 import docopt
 
 from cosmic_ray.work_db import use_db, WorkDB
@@ -9,7 +10,7 @@ from cosmic_ray.work_db import use_db, WorkDB
 def format_survival_rate():
     """cr-rate
 
-Usage: cr-rate [--estimate] [--confidence <confidence_rate>] <session-file>
+Usage: cr-rate [--estimate] [--confidence <confidence_rate>] [--fail-over <max_value>] <session-file>
 
 Calculate the survival rate of a session.
 
@@ -19,11 +20,16 @@ options:
     --confidence <confidence_rate> Specify the confidence levels for estimates, 95 (%) by
                                    default. 80, 90, 95, 98, 99, 99.5, 99.8 and 99.9 are
                                    supported.
+    --fail-over <max_value>        Exit with a non-zero code if the survival rate is
+                                   larger than <max_value> or the calculated confidence
+                                   interval is above the <max_value> (if --estimate is used).
+                                   Specified as percentage.
 """
     arguments = docopt.docopt(
         format_survival_rate.__doc__, version='cr-rate 1.0')
     show_estimate = arguments['--estimate']
     confidence = arguments['--confidence']
+    fail_over = arguments['--fail-over']
 
     # use integers as keys as equality is not well defined for floats
     # the values are z-values for Standard Normal Probabilities
@@ -53,10 +59,16 @@ options:
     if show_estimate:
         conf_int = math.sqrt(rate * (100-rate) / num_complete) \
                    * z_score * (1 - math.sqrt(num_complete / num_items))
+        min_rate = rate - conf_int
         print('{:.2f} {:.2f} {:.2f}'
-              .format(rate-conf_int, rate, rate+conf_int))
+              .format(min_rate, rate, rate + conf_int))
+
     else:
         print('{:.2f}'.format(rate))
+        min_rate = rate
+
+    if fail_over and min_rate > float(fail_over):
+        sys.exit(1)
 
 
 def survival_rate(work_db):
--- a/src/cosmic_ray/tools/survival_rate.py
+++ b/src/cosmic_ray/tools/survival_rate.py
@@ -9,18 +9,41 @@ from cosmic_ray.work_db import use_db, WorkDB
 def format_survival_rate():
     """cr-rate
 
-    Usage: cr-rate --estimate <session-file>
+Usage: cr-rate [--estimate] [--confidence <confidence_rate>] <session-file>
 
-    Calculate the survival rate of a session.
+Calculate the survival rate of a session.
 
-    options:
-        --estimate    Print the lower bound, estimate and upper bound of
-                      survival rate
-    """
+options:
+    --estimate                     Print the lower bound, estimate and upper bound of
+                                   survival rate
+    --confidence <confidence_rate> Specify the confidence levels for estimates, 95 (%) by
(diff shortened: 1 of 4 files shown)


abingham:
I need to give this some thought. I'm very reluctant to modify the work-db in this way. This modification only makes sense in the context of this kind of sampling approach, and I want to avoid the temptation to change core parts of CR for features that can be handled in other, less intrusive ways, less tightly-coupled ways.

At the same time, I really like the idea of this feature, and I want to enable it. And there's an elegance to the randomization approach you've implemented. I just need to toss it around in my head a bit.

My inclination right now is to merge your change in with the understanding that I might change it later. I've got plans to rework the interceptor system into something less tightly coupled with the `init` command, and once I do that I'll look at reimplementing the sampling approach with an interceptor (or whatever I change that name to...it's a goofy name!) How does that sound?

tomato42:
So, I went for this approach precisely because it doesn't require any special setup, that means that if you decide after an hour (or 2, or 12) that you can't wait any more, you can just kill `exec` and even the regular `cr-report` will report survival rate that is closer to the real one than with regular run that goes module by module. So it's making every tool a better estimator this way, even without including the `cr-rate` change.

But when running in CI you want to minimise false positives, so that's why I added the `cr-rate` changes.

If you'd prefer a special option to `exec` that will run the tests in random, not orderly fashion, I can do that too (making `pending_work_items` copy that uses the randomisation would be quite trivial). But then why one would depend on running the tests module by module? (and not run the test against given module only...) Am I missing something?

Side note: I have used this code in integration with travis: https://github.com/warner/python-ecdsa/pull/156 and it doesn't look _too_ gnarly (old, completed run is [here](https://travis-ci.org/warner/python-ecdsa/builds/606525550?utm_medium=notification&utm_source=github_status)), the only downside is that in 20 minutes I can test just around 20 mutations (but that's because the code is slow in general). The real survival rate for this repo is 23.93%.

abingham:
> So it's making every tool a better estimator this way,

Yes, this is an aspect of you approach that I particularly like.

> But then why one would depend on running the tests module by module?

This is something I was thinking over last night, and I can't think of any reason why users would (or should) depend on testing order. If it turns out that I'm wrong then we can add some sort of "test in discovery order" flag, with random order being the default behavior.

I've come around to thinking that your change is a good way to go, and I'm happy to implement the randomization in the work-db. With that said, this randomization behavior is now critical for certain features of CR and we should have some sort of test for it. Can you add some kind of test for the randomization? Even something that just runs `pending_work_items` twice and checks that the orders are different would be fine (though even this is technically insufficient...testing for randomness is awkward).

If you'd prefer, I can merge this in now as-is and you can write the test as a second PR. Or I can wait for you to add it to this PR. I'm going to be traveling for the next week or so, so I may not be as responsive to PRs as normal.

tomato42:
> With that said, this randomization behavior is now critical for certain features of CR and we should have some sort of test for it.

good point, will add something for that today

tomato42:
@abingham ok, unit test case to verify that the query returns random orders added

src/cosmic_ray/tools/survival_rate.py
```diff
@@ -8,16 +10,65 @@
...
+    supp_z_scores = {800: 1.282,
+                     900: 1.645,
+                     950: 1.960,
+                     980: 2.326,
+                     990: 2.576,
+                     995: 2.807,
+                     998: 3.080,
+                     999: 3.291}
+
+    if not confidence:
```
abingham: I know a confidence of zero is unlikely, but if someone supplied it for some reason then this would set it to 95. Should this check be `if confidence is None`?

src/cosmic_ray/tools/survival_rate.py
```diff
@@ -8,16 +10,65 @@
...
+                                   interval is above the <max_value> (if --estimate is used).
+                                   Specified as percentage.
+"""
     arguments = docopt.docopt(
         format_survival_rate.__doc__, version='cr-rate 1.0')
+    show_estimate = arguments['--estimate']
+    confidence = arguments['--confidence']
+    fail_over = arguments['--fail-over']
+
+    # use integers as keys as equality is not well defined for floats
```
abingham: I'm a bit wary of this; floating point equality is perfectly well defined, though I suspect you're (reasonably) concerned about rounding causing confusion. If we can't use floats and can't calculate z-values for arbitrary confidence intervals, would an enum be more appropriate?

src/cosmic_ray/tools/survival_rate.py
tomato42: `confidence` here is a string, isn't it? so a "0" would evaluate to True

an empty string would cause it to default to 95, and I think that's acceptable

src/cosmic_ray/tools/survival_rate.py
tomato42: > I'm a bit wary of this; floating point equality is perfectly well defined,

hmm, since we're not calculating it, but just comparing hard-coded values, yes, equality would work

> If we can't use floats and can't calculate z-values for arbitrary confidence intervals

sure, it is possible to calculate it, but as far as I can tell, it would require integration of a not exactly trivial function so I don't think it's really worth it

> would an enum be more appropriate?

if the confidence intervals calculation was to be supported for other values, and the code generalised, probably; I do think that `dict()` aligns better with a "lookup table" than an enum though, and in the future we may need to add 99.99 or some-such

PR #495, opened 2020-01-30
Add git-branch config

ptitpoulpe:
Add a new 'git-branch' config.
If a branch is given, new code in modules will be mutated

### DIFF
--- a/docs/source/filters.rst
+++ b/docs/source/filters.rst
@@ -64,6 +64,17 @@ cr-filter-pragma
 The ``cr-filter-pragma`` filter looks for lines in your source code containing the comment "# pragma: no mutate". Any
 mutation in a session that would mutate such a line is skipped.
 
+cr-filter-git
+-------------
+
+The ``cr-filter-git`` filter looks for edited or new lines from the given git branch. Any mutation in a session that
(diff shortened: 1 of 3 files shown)


abingham:
It looks like this could be implemented as filter:

https://cosmic-ray.readthedocs.io/en/latest/filters.html

If so (and if there's no strong reason not to), then that's how this should be done. I think we'll find that an extrinsic, filter-based approach is more powerful anyway...we can grow its complexity arbitrarily without needing to touch the CR core.

ptitpoulpe:
ok I will do it like as a filter, but it could be faster to put filter in init to filter before mutation.

abingham:
> but it could be faster to put filter in init to filter before mutation.

Mutation only happens in the `exec` phase. Since you'll (presumably) run your filter after `init` but before `exec`, you will be doing filtering before mutation.

So while your `init` + filter phase might be slightly longer, I don't think you'll see any substantial speed problems overall.

ptitpoulpe:
Make it a filter, will update doc

ptitpoulpe:
Doc added

abingham:
This all look great, thanks!

PR #502, opened 2020-07-15
Issue 501 - Show Survived Mutant Counts in Reports

XD-DENG:
Without this patch, in both CLI and HTML result, we would only have 'Survival rate', while people may be interested to know the count of survived mutants as well.

### After Applying This Patch

Screenshot_2020-07-15_at_9_41_42_PM

<hr>

Screenshot_2020-07-15_at_9_41_54_PM

### DIFF
--- a/src/cosmic_ray/tools/html.py
+++ b/src/cosmic_ray/tools/html.py
@@ -9,7 +9,7 @@ from yattag import Doc
 
 from cosmic_ray.work_db import WorkDB, use_db
 from cosmic_ray.work_item import TestOutcome
-from cosmic_ray.tools.survival_rate import survival_rate
+from cosmic_ray.tools.survival_rate import kills_count, survival_rate
 
 
 def report_html():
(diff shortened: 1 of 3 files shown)


src/cosmic_ray/tools/html.py
```diff
@@ -89,7 +89,8 @@ def _generate_html_report(db, only_completed, skip_success):
                                         text('Complete: {} ({:.2f}%)'.format(
                                             num_complete, num_complete / num_items * 100))
                                     with tag('p'):
-                                        text('Survival rate: {:.2f}%'.format(survival_rate(db)))
+                                        num_killed = kills_count(db)
+                                        text('Survived mutants: {} ({:.2f}%)'.format(num_complete - num_killed, survival_rate(db)))
```
abingham, line 93: Change this to "Surviving mutants".

src/cosmic_ray/tools/survival_rate.py
```diff
@@ -71,10 +71,14 @@ def format_survival_rate():
         sys.exit(1)
 
 
+def kills_count(work_db):
```
abingham, line 74: Add a docstring to this function.

abingham:
Thanks, this looks really good! I just had a few small notes. Once those are done, I'll be happy to merge this in.

src/cosmic_ray/tools/html.py
XD-DENG: Addressed in both `html.py` and `report.py`

src/cosmic_ray/tools/survival_rate.py
XD-DENG: Addressed.

abingham:
Thanks!

XD-DENG:
Thanks for merging ;)

PR #503, opened 2020-07-17
BugFix: empty 'exclude_operators' results in skipping all operators

XD-DENG:
WithOUT this patch, if a user (by accident) provides an empty `exclude-operators` in the configuration then applies command `cr-filter-operators`, it will results in that ALL operators to be skipped, while it should not skip any operators.

It's easy to check & confirm this bug:

```python
import re
exclude_operators = []
re_exclude_operators = re.compile('|'.join('(:?%s)' % e for e in exclude_operators))
assert re_exclude_operators.match("abc")
assert re_exclude_operators.match("xyz")
```

Tests are updated/added.


abingham:
Thanks!

Issue #525, opened 2022-02-07
Excluded modules does not work because it's not referenced correctly in cli.py


Jamie-McKernan:
As title says, if you look at the code here:
https://github.com/sixty-north/cosmic-ray/blob/cf7fb7c3cc564db1e8d53b8e8848c9f46e61a879/src/cosmic_ray/cli.py#L89

they reference `exclude-modules` in the config, yet everywhere else (including the docs) refers to this as `excluded-modules`.

I propose that the quoted line is changed to `excluded-modules` to match where it is used everywhere else. When you init a new TOML file it is created as `excluded-modules`. See examples here:

https://github.com/sixty-north/cosmic-ray/blob/ec86f49a5d48b6e7de656e3674eeaa971ec1c7b6/src/cosmic_ray/commands/new_config.py#L44

https://github.com/sixty-north/cosmic-ray/blob/ec86f49a5d48b6e7de656e3674eeaa971ec1c7b6/docs/source/tutorials/intro/tutorial.toml.1#L4

As a side note to this, I think the docs could be improved with some examples of excluding modules, as it's not clear that cosmic-ray is expecting glob patterns. I think something like using the example of `**/*_test.py` could be a good way of excluding the tests themselves from being mutated (something I had to do recently).

I'm happy to fork and make the change when I find time, but thought I'd post the issue here before doing so.

abingham:
Wow, that's definitely a bug! Yes, changing that code in cli.py to say 'excluded-modules' is certainly the right thing do to. Better examples in the docs would be very helpful as well. If you can put those changes together, I'll be happy to merge them in. Thanks!

Jamie-McKernan:
@abingham PR has been made https://github.com/sixty-north/cosmic-ray/pull/526

PR #526, opened 2022-02-16
Issue #525 - Fixing excluded-modules

Jamie-McKernan:
Updated the cli.py file to have the correct reference to excluded-modules and updated the docs to include a brief example of how to use the excluded-modules.

Solves https://github.com/sixty-north/cosmic-ray/issues/525


abingham:
Thanks!

Issue #528, opened 2022-08-22
Is it possible to define a mutation operator that takes arguments?


AndrewC19:
I would like to define a custom mutation operator that looks for particular variables in the program-under-test and replaces them with random constants. For example, I might want to replace all uses of a variable named `x1` with some randomly sampled int.

Is there a straightforward way to achieve this using cosmic-ray? I have tried defining my own `Operator` sub-class but I need a way to pass the target variable to the mutation operator.

Thanks in advance.

abingham:
There is not currently a way to do that. I guess broadly speaking you'd want to be able to do some per-operator configuration in the configuration file and have it take effect when you exec your session. This seems totally reasonable. A few things come to mind which we'd need to think about.

1. How will this argument information get communicated through distributors? Right now, the WorkItems that we send over the distributor don't have configuration information in them, so the workers - which could be running anywhere - don't know about it. We could send the complete configuration over to each worker somehow, or - more directly addressing your issue, and I think my preference - each WorkItem could also include some kind of "arguments" struct that the operator could interpret as it sees fit. I don't think this would be too hard, really.

2. Do we need to support multiple sets of arguments for any given operators? That is, suppose you wanted to replace all variables named `x1` *and* you wanted to replace all  variables named `x2`? Do operators now become templates (in the C++ sense, sorta) where each set of arguments actually generates a new independent operator? Something like this could possibly be handled by our `OperatorProvider` system, but the providers would probably need to be handed the configuration. (and perhaps I'm just overthinking this).

There are probably other angles to this that I'm not seeing right now. Does any of this sound reasonable to you, or along the lines of what you had in mind? I don't know when I'd be able to work on this, but it feels like a good "airplane project"...I just need a long trip somewhere!

AndrewC19:
Thanks for the quick response @abingham.

Both of these solutions sound sensible to me but perhaps some additional info will help to narrow it down.

Ultimately, I want to be able to make three kinds of mutation that target specific lines of code based on their variable usage. Here I will use the term 'used' to mean that a variable appears on the right hand side of `=` and 'declared' for variables on the left hand side (e.g. in `y = 2*x1`, `y` is declared and `x1` is used). 

1. Find lines of code where a variable, say `x1`, is used and replace `x1` with a constant (e.g. `y1 = 2*x1` --> `y1 = 2*5`).
[...]

PR #529, opened 2022-08-25
Mutation operators with arguments (Issue #528)

AndrewC19:
### Overview
This PR enables users to define mutation operators with arbitrary key word arguments. As an example, the `VariableReplacer` and `VariableInserter` operators have been added. These take two arguments, a `cause_variable: str` and an `effect_variable: str`, and enable the following three types of mutation:

(1) Replacing *all* usages of a specific variable with a random numerical constant (e.g.  `VariableReplacer('x1')` mutates `y=x1+x2+x3+10` to `y=30+x2+x3+10`).
(2) Replacing usages of a specific variable that appear in the statement of another specific variable (e.g. `VariableReplacer(x1, y)` mutates `y=x1+x2+x3+10` to `y=30+x2+x3+10` but does not mutate `j=x1+x2+x3+10`).
(3) Adding usages of a specific variable in the statement of another specific variable (e.g. `VariableInserter(x, y)` mutates `y=2*z+1` to `y=2*z+1*x`).

To specify the variables of interest, the user can now add the following table to their TOML config file:
```TOML
[[cosmic-ray.operators]]
name = "core/VariableReplacer"
args = [{ cause_variable = "x", effect_variable = "y"},
        { cause_variable = "x", effect_variable = "j"}]
 
[[cosmic-ray.operators]]
name = "core/NumberReplacer"
```
If such a table is added, only the specified mutation operators will be applied (notice that args are only supplied to operators with args). Otherwise, as before, all mutation operators are tried during initialisation. This will not apply any operators that require arguments.

### Changes
As part of this PR, the following changes have been made:

- (`src/cosmic_ray/operators/variable_replacer.py`): A new operator that replaces named variable usages, as described above.

- (`src/cosmic_ray/operators/variable_inserter.py`): A new operator that added named variable usages, as described above.

- (`src/cosmic_ray/operators/provider.py`): `provider.py` now includes `VariableReplacer` in the available operators.

- (`src/cosmic_ray/work_item.py`): `ResolvedMutationSpec` now includes a dictionary of operator arguments as an additional optional argument.

- (`src/cosmic_ray/work_db.py`): `work_db.py` now stores an additional column in the `MutationSpecStorage` table for the operator arguments. These are stored as a JSON object, enabling an arbitrary number of key word arguments to be stored and retrieved.

- (`src/cosmic_ray/config.py`): `config.py` now has an additional property `operators` that stores the specified operators and their args.

- (`src/cosmic_ray/cli.py`): If the `cosmic-ray.operators` table is present in the config file,  it is extracted and passed as a dictionary to `cosmic-ray.commands.init`. This dictionary stores the `name` and `args` (optional) of the operator.

- (`src/cosmic_ray/commands/init.py`): `init.py` now instantiates a dictionary storing all operator names (as before) *only* if the operator config hasn't been provided. Otherwise, it passes the dictionary containing the specified operators and args to `_all_work_items`. Then, `_all_work_items` handles this dictionary by instatiating an operator with any specified args and identifiying mutation positions, as before.

- (`src/cosmic_ray/mutating.py`): `mutating.py` now passes any arguments to the instantiation of the operator by unpacking the `mutation_args` dict. This is an empty dict if no args have been specified. A try except is used to handle cases where no args are given (e.g. as in `test_no_test_return_value`).

### TODOs:
The following problems still need addressing before merging the PR:

- HTML Report: Currently, the HTML report does not output information about the arguments of the `VariableReplacer` operator. It would be good if the report listed the arguments of the mutation operator if there are any.

- Tests based on `Operator.example` class method: Currently, the `test_operators.py` tests use the tuple defined in the `example` method to build an assertion. For operators that use arguments, this tuple needs to store information about the arguments used for each test (currently this information is put in a comment next to the tuple). I couldn't find an elegant solution for this problem as the `example` method is a class method, so has no access to the instance variables.

- Update `example` method of `VariableInserter`. This currently has the body of `VariableReplacer`'s `example` method.

### DIFF
--- a/src/cosmic_ray/cli.py
+++ b/src/cosmic_ray/cli.py
@@ -84,10 +84,7 @@ def init(config_file, session_file):
     executed with other commands.
     """
     cfg = load_config(config_file)
-    try:
-        operators_cfg = cfg.operators_config
-    except KeyError:
-        operators_cfg = None
+    operators_cfg = cfg.operators_config
     modules = cosmic_ray.modules.find_modules(Path(cfg["module-path"]))
     modules = cosmic_ray.modules.filter_paths(modules, cfg.get("excluded-modules", ()))
 
--- a/src/cosmic_ray/cli.py
+++ b/src/cosmic_ray/cli.py
@@ -97,7 +97,7 @@ def init(config_file, session_file):
             log.info(" - %s: %s", directory, ", ".join(sorted(files)))
 
     with use_db(session_file) as database:
-        cosmic_ray.commands.init(modules, database)
+        cosmic_ray.commands.init(modules, database, cfg.operators_config)
 
     sys.exit(ExitCode.OK)
 
--- a/src/cosmic_ray/cli.py
+++ b/src/cosmic_ray/cli.py
@@ -84,7 +84,10 @@ def init(config_file, session_file):
     executed with other commands.
     """
     cfg = load_config(config_file)
-
+    try:
+        operators_cfg = cfg.operators_config
+    except KeyError:
+        operators_cfg = None
     modules = cosmic_ray.modules.find_modules(Path(cfg["module-path"]))
     modules = cosmic_ray.modules.filter_paths(modules, cfg.get("excluded-modules", ()))
 
@@ -97,7 +100,7 @@ def init(config_file, session_file):
             log.info(" - %s: %s", directory, ", ".join(sorted(files)))
 
     with use_db(session_file) as database:
-        cosmic_ray.commands.init(modules, database, cfg.operators_config)
+        cosmic_ray.commands.init(modules, database, operators_cfg)
 
     sys.exit(ExitCode.OK)
 
--- a/src/cosmic_ray/commands/init.py
+++ b/src/cosmic_ray/commands/init.py
@@ -25,24 +25,28 @@ def _all_work_items(module_paths, operator_cfgs) -> Iterable[WorkItem]:
                 operator_args = operator_cfg["args"]
 
             for args in operator_args:
-                operator = get_operator(operator_name)(**args)
-
-                positions = (
-                    (start_pos, end_pos)
-                    for node in ast_nodes(module_ast)
-                    for start_pos, end_pos in operator.mutation_positions(node)
-                )
-
-                for occurrence, (start_pos, end_pos) in enumerate(positions):
-                    mutation = ResolvedMutationSpec(
-                        module_path=str(module_path),
-                        operator_name=operator_name,
-                        operator_args=args,
-                        occurrence=occurrence,
-                        start_pos=start_pos,
-                        end_pos=end_pos,
+                try:
+                    operator = get_operator(operator_name)(**args)
+                except TypeError:
+                    # Operator can't be loaded because it requires args, try next type
+                    continue
+                else:
+                    positions = (
+                        (start_pos, end_pos)
+                        for node in ast_nodes(module_ast)
+                        for start_pos, end_pos in operator.mutation_positions(node)
                     )
-                    yield WorkItem.single(job_id=uuid.uuid4().hex, mutation=mutation)
+
+                    for occurrence, (start_pos, end_pos) in enumerate(positions):
+                        mutation = ResolvedMutationSpec(
+                            module_path=str(module_path),
+                            operator_name=operator_name,
+                            operator_args=args,
+                            occurrence=occurrence,
+                            start_pos=start_pos,
+                            end_pos=end_pos,
+                        )
+                        yield WorkItem.single(job_id=uuid.uuid4().hex, mutation=mutation)
 
 
 def init(module_paths, work_db: WorkDB, operators_cfgs=None):
--- a/src/cosmic_ray/commands/init.py
+++ b/src/cosmic_ray/commands/init.py
@@ -8,37 +8,51 @@ import cosmic_ray.modules
 from cosmic_ray.work_item import MutationSpec, ResolvedMutationSpec, WorkItem
 from cosmic_ray.plugins import get_operator
 from cosmic_ray.work_db import WorkDB
+from cosmic_ray.operators.variable_replacer import VariableReplacer
 
 log = logging.getLogger()
 
 
-def _all_work_items(module_paths, operator_names) -> Iterable[WorkItem]:
+def _all_work_items(module_paths, operator_cfgs) -> Iterable[WorkItem]:
     "Iterable of all WorkItems for the given inputs."
+
     for module_path in module_paths:
         module_ast = get_ast(module_path)
 
-        for op_name in operator_names:
-            operator = get_operator(op_name)()
+        for operator_cfg in operator_cfgs:
+            operator_name = operator_cfg["name"]
+            if "args" not in operator_cfg:
+                operator_args = [{}]
+            else:
+                operator_args = operator_cfg["args"]
 
-            positions = (
-                (start_pos, end_pos)
-                for node in ast_nodes(module_ast)
-                for start_pos, end_pos in operator.mutation_positions(node)
-            )
+            for args in operator_args:
+                operator = get_operator(operator_name)(**args)
+                # if op_name == "core/VariableReplacer":
+                #     operator = VariableReplacer(cause_variable="x")
+                # else:
+                #     operator = get_operator(op_name)(**operator_args)
 
-            for occurrence, (start_pos, end_pos) in enumerate(positions):
-                mutation = ResolvedMutationSpec(
-                    module_path=str(module_path),
-                    operator_name=op_name,
-                    occurrence=occurrence,
-                    start_pos=start_pos,
-                    end_pos=end_pos,
+                positions = (
+                    (start_pos, end_pos)
+                    for node in ast_nodes(module_ast)
+                    for start_pos, end_pos in operator.mutation_positions(node)
                 )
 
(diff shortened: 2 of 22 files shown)


abingham:
This is really excellent work! I want take some time this weekend to look/think it over. The spirit of it seems absolutely correct, and I suspect we'll just take the change wholesale.

AndrewC19:
Thanks, that sounds good to me. Happy to make changes to improve the quality of the changes when I get the time.

abingham:
> For operators that use arguments, this tuple needs to store information about the arguments used for each test (currently this information is put in a comment next to the tuple). I couldn't find an elegant solution for this problem as the example method is a class method, so has no access to the instance variables.

It seems to me that the `examples()` classmethods could simply include any necessary arguments in the returned tuple (as a side note, we should probably return something more structured from those functions, e.g. a dataclass). So the examples for `VariableReplacer` might be:
```
@classmethod
    def examples(cls):
        return (
            ('y = x + z', 'y = 10 + z', 0, {'cause_variable': 'x'}),
            ('j = x + z\ny = x + z', 'j = x + z\ny = -2 + z', 0, {'cause_variable': 'x', 'effect_variable': 'y'}),
            # etc...
        )
```
Those could be **-unpacked into the constructor call in the test. Does make sense? Am I missing some angle?

src/cosmic_ray/cli.py
```diff
@@ -84,7 +84,10 @@ def init(config_file, session_file):
     executed with other commands.
     """
     cfg = load_config(config_file)
-
+    try:
```
abingham, line 87: I think this check is better handled in `load_config()` itself. I.e. `load_config()` should always return a config with a `operators_config` field, even if the value is `None`. That way we don't have to duplicate this logic in other (potential) places where `load_config()` is called.

src/cosmic_ray/commands/init.py
```diff
@@ -2,43 +2,54 @@
...
     "Iterable of all WorkItems for the given inputs."
+
     for module_path in module_paths:
         module_ast = get_ast(module_path)
 
-        for op_name in operator_names:
-            operator = get_operator(op_name)()
+        for operator_cfg in operator_cfgs:
+            operator_name = operator_cfg["name"]
+            if "args" not in operator_cfg:
```
abingham, line 22: I think this conditional would be better expressed as `operator_cfg.get('args', [{}])`

abingham: that is, it would be clearer to readers.

src/cosmic_ray/commands/init.py
```diff
@@ -2,43 +2,54 @@
...
+            else:
+                operator_args = operator_cfg["args"]
 
-            positions = (
-                (start_pos, end_pos)
-                for node in ast_nodes(module_ast)
-                for start_pos, end_pos in operator.mutation_positions(node)
-            )
+            for args in operator_args:
+                try:
```
abingham, line 28: This concerns me a bit. If there's a typo in an argument name in the config, we'll silently ignore that here by assuming the resulting TypeError means that the operator didn't get *any* args. A typo like this should result in an error, I think.

I'd be more comfortable doing something like this. If there's a TypeError and `args` is empty, we skip the operator. If `args` is *not* empty, though, we should raise an exception and halt initialization. This way we can let the user know that they've made what appears to be a mistake.

src/cosmic_ray/config.py
```diff
@@ -87,6 +87,11 @@ def distributor_config(self):
         name = self.distributor_name
         return self["distributor"].get(name, ConfigDict())
 
+    @property
```
abingham, line 90: I mentioned elsewhere that I think this should always return a value. If we ultimately decide, however, to leave this method as-is, it needs a docstring explaining that it might raise `KeyError`. Currently, that would come as a big surprise to some callers.

src/cosmic_ray/mutating.py
```diff
@@ -50,7 +49,12 @@ async def mutate_and_test(mutations: Iterable[MutationSpec], test_command, timeo
             file_changes = {}
             for mutation in mutations:
                 operator_class = cosmic_ray.plugins.get_operator(mutation.operator_name)
-                operator = operator_class()
+                try:
+                    operator_args = mutation.operator_args
```
abingham, line 53: If `mutation` truly is a `ResolvedMutationSpec` as the annotation indicates, it should always have an `operator_args` attribute. This `AttributeError` check seems incorrect.

I think that really the `operator_args` attribute should be on `MutationSpec`, not `ResolvedMutationSpec`. The `Resolved` class only exists to indicate that the position of the mutation in code has been "resolved"...the existence of this class may be a design blunder on my part.

abingham: In any event, this `AttributeError` check is probably not the right thing to do.

src/cosmic_ray/work_item.py
```diff
@@ -74,6 +74,7 @@ class ResolvedMutationSpec(MutationSpec):
     "A MutationSpec with the location of the mutation resolved."
     start_pos: Tuple[int, int]
     end_pos: Tuple[int, int]
+    operator_args: Optional[Dict] = None
```
abingham, line 77: As I mentioned elsewhere, I think this field should be on `MutationSpec`, not `ResolvedMutationSpec`.

abingham (commented):
I also think the HTTP distributor needs to be considered in light of operator arguments. Are the arguments making it across the HTTP call, and are they being used? I know we don't have any tests for this, so it's easy to miss this.

abingham:
I'm broadly happy with the direction of your changes. I think it needs a little work in some areas, we'll get operator arguments in there one way or the other.  Take a look at the review notes I made and let me know if you have any questions. It's *entirely* possible that I've missed the point of some things you've done, so let me know if that's the case.

src/cosmic_ray/mutating.py
AndrewC19: I initially tried to add `operator_args` to `MutationSpec` as an optional argument but because `ResolvedMutationSpec` has additional positional arguments, this was not possible. Any ideas how we could address this?

AndrewC19:
> I also think the HTTP distributor needs to be considered in light of operator arguments. Are the arguments making it across the HTTP call, and are they being used? I know we don't have any tests for this, so it's easy to miss this.

I don't think so. If you can direct me to the HTTP call then I can implement the necessary changes.

src/cosmic_ray/commands/init.py
AndrewC19: Good point. I hadn't thought of this corner case.

Your solution makes sense; I'll implement the changes soon.

abingham:
> > I also think the HTTP distributor needs to be considered in light of operator arguments. Are the arguments making it across the HTTP call, and are they being used? I know we don't have any tests for this, so it's easy to miss this.
> 
> I don't think so. If you can direct me to the HTTP call then I can implement the necessary changes.

That's in cosmic_ray/distribution/http.py

AndrewC19:
> > For operators that use arguments, this tuple needs to store information about the arguments used for each test (currently this information is put in a comment next to the tuple). I couldn't find an elegant solution for this problem as the example method is a class method, so has no access to the instance variables.
> 
> It seems to me that the `examples()` classmethods could simply include any necessary arguments in the returned tuple (as a side note, we should probably return something more structured from those functions, e.g. a dataclass). So the examples for `VariableReplacer` might be:
> 
> ```
> @classmethod
>     def examples(cls):
>         return (
>             ('y = x + z', 'y = 10 + z', 0, {'cause_variable': 'x'}),
>             ('j = x + z\ny = x + z', 'j = x + z\ny = -2 + z', 0, {'cause_variable': 'x', 'effect_variable': 'y'}),
>             # etc...
>         )
> ```
> 
> Those could be **-unpacked into the constructor call in the test. Does make sense? Am I missing some angle?

I have implemented an `Example` dataclass which stores the pre and post mutation code (mandatory), occurrence index (optional), and mutation args (optional). This is a more consistent solution than relying on a tuple, in my opinion. All existing operators examples have been updated to use this structure and their tests pass.

On this topic, I have encountered an issue with the operator tests for `VariableRemover` and `VariableInserter`: both randomly sample a value for the mutation (e.g. a random numerical value) and therefore I cannot say the precise value that a given mutation will have (since it is random). One solution would be to set a random seed for the tests; what are your thoughts?

abingham:
> One solution would be to set a seed for the tests; what are your thoughts?

This is an interesting topic I haven't thought about much. I'm not sure about the desirability of randomness in mutation operators, but I guess CR has to accept the possibility since it can be extended. Fixing the seed is probably a good first attempt...it should at least allow us to get the tests passing. Another option is that we allow Examples to specify a custom 'comparison' function when simple string comparison won't be enough. I need to think about this.

abingham:
I think what I'm going to do is merge this in as-is, and then I'll make some changes I want. My biggest concern at this point is that providing arguments for any operator essentially turns off the other non-argument-having operators. This seems incorrect to me. If we're going to have general support for arguments, I see no good reason to make such a big mode switch when any arguments are supplied; providing arguments and disabling operators seem like orthogonal concerns to me.

So I'm going to merge this in and make some changes afterward. If you want to be able to run *only* a subset of operators, I think we have existing features for this. If not, you can write a simple filter to do this.

AndrewC19:
Great, that sounds good to me.

There are a few issues that still need fixing from this PR:
- Adding a seed to tests for operators that include randomness.
- Setting up operator args with HTTP.

Hope that helps!

PR #531, opened 2022-11-25
Make except statment lists work with just one item

ThunderKey:
Without this patch "except (OSError):" leads to an AttributeError. This patch handles this specific example where test_list is an instance of Name.

The tests that already fail (probably since e02b8ec484f00757271e41ac1112a1a599559cb2) still fail, but the added test succeeds.


abingham:
Thanks!

PR #533, opened 2022-12-21
Allow module-path configuration to contain multiple paths

ThunderKey:
I had to estimate how many mutants survive in a huge project where around 400'000 mutations were found. For this I've randomly selected tests and only wanted to mutate the targets of those tests. With the `excluded-modules` this would have been a lot of work. For that I've adapted the code to allow `module-path` to be a list of paths, which will be used as mutation targets.

Since I've already done the work I thought you might be interested in it. If not feel free to discard this PR :)

Additionally, I would expect the tool to abort if the `module-path` is not found and not just do nothing. For an example see: https://github.com/sixty-north/cosmic-ray/blob/e5e0e3b4626da101975fa9ce2be3ce50966c97d2/tests/unittests/test_find_modules.py#L34-L36

Since this changes the behaviour I didn't want to just include it, but if you agree I can also add a commit that checks that the file/folder exists.

### DIFF
--- a/docs/source/tutorials/intro/index.rst
+++ b/docs/source/tutorials/intro/index.rst
@@ -136,6 +136,8 @@ refers to a single top-level module that will be mutated, and in this case we're
 
     When working on a package, Cosmic Ray will apply mutations to all submodules in the package.
 
+    Additionally, the 'module-path' can be a list of directories or files: `module-path = ["file1.py", "some_directory"]`
+
 Line 3 tells Cosmic Ray the maximium amount of time to let a test run before it's considered a failure:
 
 .. literalinclude:: tutorial.toml.1
--- a/src/cosmic_ray/cli.py
+++ b/src/cosmic_ray/cli.py
@@ -85,7 +85,9 @@ def init(config_file, session_file):
     """
     cfg = load_config(config_file)
     operators_cfg = cfg.operators_config
-    modules = cosmic_ray.modules.find_modules(Path(cfg["module-path"]))
+
+    module_paths = [Path(cfg["module-path"])] if isinstance(cfg["module-path"], str) else map(Path, cfg["module-path"])
+    modules = cosmic_ray.modules.find_modules(module_paths)
     modules = cosmic_ray.modules.filter_paths(modules, cfg.get("excluded-modules", ()))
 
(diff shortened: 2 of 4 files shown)


abingham:
This is great, thanks!

> Additionally, I would expect the tool to abort if the module-path is not found and not just do nothing.

This is probably better behavior than we have now. So if you've got a patch that does this, I'd be happy to consider it.

> I had to estimate how many mutants survive in a huge project where around 400'000 mutations were found. For this I've randomly selected tests and only wanted to mutate the targets of those tests.

This is an interesting topic. Another way to approach it might be to write a filter that skips all but a random sampling of mutations, letting the user specify, say, a percentage of tests to run or something.

ThunderKey:
> This is probably better behavior than we have now. So if you've got a patch that does this, I'd be happy to consider it.

I've created #534 with a patch to verify that the paths exist.

> This is an interesting topic. Another way to approach it might be to write a filter that skips all but a random sampling of mutations, letting the user specify, say, a percentage of tests to run or something.

That would also be an interesting approach. I guess the difficulty would be to detect which mutations belong to which test, without skipping too much that is not covered. Otherwise the resulting score will not be a reliable estimate.

abingham:
> I've created https://github.com/sixty-north/cosmic-ray/pull/534 with a patch to verify that the paths exist.

Thanks!

> I guess the difficulty would be to detect which mutations belong to which test, without skipping too much that is not covered. Otherwise the resulting score will not be a reliable estimate.

Ah, I hadn't read your initial problem description closely enough. I don't tend to think of mutations as belonging to a particular test since, in general, mutations can change program behavior drastically enough that all sorts of "unrelated" tests might be affected. With that said, your approach of using the relationship between modules and their tests as a proxy is a practical one most of the time.

I think you could design the filter to a) skip all mutations in modules you're not interested in at all and then b) randomly skip those in modules of interest until you've driven the number of mutations to the target.

The only reason I belabor this idea at all is that I've learned to be reluctant to make changes to the "core" of CR when the work can be done with a filter program. I'm talking myself through my anxiety about making the core any more complex than it needs to be 😅

ThunderKey:
That's a very nice idea with the filters. I'll have to read up more about them :)

This totally makes sense to keep the core as simple as it can be.

PR #534, opened 2022-12-27
Verify that the module paths exist

ThunderKey:
As mentioned in #533, I'm providing a patch so that the module paths are verified to exist.

An example output of an init command where the module path (`some_inexisting_file` in this case) does not exist:

```bash
$ cosmic-ray init cosmic-ray.toml cosmic-ray.sqlite                                                        
Could not find module path some_inexisting_file
```


abingham:
Thanks!

PR #540, opened 2023-12-29
process work items in random order

tomato42:
for the estimation of the survival rate to be representative, the sample must be random, so execute the tasks in random order

see #490 and #491


abingham:
This looks good, and I can't see how it should cause problems for anyone. If we find that it does cause problems, i.e. if some people really do need mutations processed in some order, then we'll deal with that later. Thanks!

PR #541, opened 2023-12-29
cr-filter-git: mark all as skipped in single transaction

tomato42:
Because every skipped transaction is committed individually to the DB, it takes extensive amount of time for a database with few tens of thousands of mutants:

without the patch:
```
18.08user 3.50system 1:12.85elapsed 29%CPU (0avgtext+0avgdata 116220maxresident)k
0inputs+1014816outputs (0major+27993minor)pagefaults 0swaps
```

with this patch:
```
10.33user 0.18system 0:10.55elapsed 99%CPU (0avgtext+0avgdata 116008maxresident)k
0inputs+3176outputs (0major+27989minor)pagefaults 0swaps
```

And this is with a fairly fast NVME SSD, slower disk storage will take extensively more time.

### DIFF
--- a/src/cosmic_ray/tools/filters/git.py
+++ b/src/cosmic_ray/tools/filters/git.py
@@ -45,6 +45,8 @@ class GitFilter(FilterApp):
     def _skip_filtered(self, work_db, branch):
         git_news = self._git_news(branch)
 
+        job_ids = []
+
         for item in work_db.pending_work_items:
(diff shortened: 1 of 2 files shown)


abingham:
This sounds great. When I'm back from vacation I'll look it over and likely merge it right in. Thanks!

Sent from ProtonMail mobile

-------- Original Message --------

PR #545, opened 2024-04-08
cr-rate: don't crash with no executed and --estimate specified

tomato42:
When using the `cr-rate` in CI, and running tests only for the changed lines, when a PR changes only documentation, the `cr-rate` tool will crash when it's used with the `--estimate` option.

This fixes the issue.


abingham:
Thanks!

PR #551, opened 2024-10-13
Enables ruff UP rule

jairhenrique:


abingham:
Will these linter rules suggest things which aren’t valid in all of our supported Python versions?

jairhenrique:
> Will these linter rules suggest things which aren’t valid in all of our supported Python versions?

@abingham no, UP looks to requires-python on pyproject.toml to determine minimum version of support new pythons features.

abingham:
Excellent, thanks for clarifying.

PR #553, opened 2024-10-13
Fixes README python badges

jairhenrique:


abingham:
Thanks! It’s really helpful to get these kinds of updates.

PR #557, opened 2024-10-13
Drops Python 3.8 support

jairhenrique:
Python 3.8 had end of life in 07 Oct 2024.

https://endoflife.date/python


abingham:
I'm a bit hesitant to remove this since I don't have much data on who is using CR and what versions they require. But since it's a relatively small change and easy to reverse, we'll merge it in any see if there's any response.

Issue #417, opened 2018-12-19
Init should refuse if there are existing results


abingham:
The init command should refuse to do anything if it sees any results in the session. We can provide a flag to the command that overrides this.

muhd360:
@abingham can you pls link the respective file in this regards???

abingham:
The command-line tool for `init` is implemented in `cosmic_ray.cli.init()`. I *think* all that we need to do is add a check if `database.num_results > 0`, refusing to continue if that's true. We could also add a command line option to allow initialization even in the presence of existing results, e.g.:
```
if database.num_results == 0 or force:
    cosmic_ray.commands.init(modules, database, operators_cfg)
``` 
Something like that.

muhd360:
ok ill make the necessary changes and raise a pr can you pls assign the hacktoberfest label to this issue??

muhd360:
hi @abingham a few follow up questions.
1.what is the test file in this scenario??
2. I understand --force will be a command in the cli .Since init expects only 2 args how can we pass force as a param???

abingham:
1. You should add tests to tests/unittests/test_command_line_processing.py. We should have a test for both the positive and negative case, i.e. the existence or not of results.

2. You'll need to modify cli.init() to accept a 'force' argument, and add a new click option decorator for 'force'.

muhd360:
hi @abingham im having build issues after using pyproject.toml to install the files and dependencies.Have i missed something????

abingham:
You’ll need to give me more details than that. What’s failing?

muhd360:
hi, i cant import the functions or the requirements.Its simply coming as module not found.I tried re-installing my virtual enviornment 2 times but still the issue wasn't fixed
![image](https://github.com/user-attachments/assets/aae3cd77-1538-4da2-b53e-42f12443112f)

abingham:
I can only guess that your IDE is looking at a different virtual environment than the one you installed into. Can you check that they're the same environment?

abingham:
Also, check that the pytest you're running is in the same venv as the code. If not, you might see similar errors. If you're not using uv to install Cosmic Ray, then you'll have to make sure to install pytest correctly yourself.

muhd360:
what is uv??? forgive my ignorance

muhd360:
hi i got it it was a vs code issue

PR #560, opened 2024-10-28
trying to fix-Init should refuse if there are existing results

muhd360:
issue number -#417
changes:-added force flag
added tests

### DIFF
--- a/src/cosmic_ray/cli.py
+++ b/src/cosmic_ray/cli.py
@@ -63,7 +63,7 @@ click.argument()
 
 
 @cli.command()
-@click.argument("force", type=bool, default=False)
+@click.option("--force", is_flag=True, default=False, help="Force initialization")
 @click.argument("config_file")
 @click.argument(
     "session_file",
@@ -97,10 +97,12 @@ def init(config_file, session_file,force):
             log.info(" - %s: %s", directory, ", ".join(sorted(files)))
 
     with use_db(session_file) as database:
-        if database.num_results>0:
-            sys.exit(ExitCode.OK)
-        elif database.num_results==0 or force:
-            cosmic_ray.commands.init(modules, database, operators_cfg)
+
+        if database.num_results>0 or not force:
+            return
+
+        
+        cosmic_ray.commands.init(modules, database, operators_cfg)
             
 
     sys.exit(ExitCode.OK)
--- a/src/cosmic_ray/cli.py
+++ b/src/cosmic_ray/cli.py
@@ -63,12 +63,13 @@ click.argument()
 
 
 @cli.command()
+@click.argument("force", type=bool, default=False)
 @click.argument("config_file")
 @click.argument(
     "session_file",
     # help="The filename for the database in which the work order will be stored."
 )
-def init(config_file, session_file):
+def init(config_file, session_file,force):
     """Initialize a mutation testing session from a configuration. This
     primarily creates a session - a database of "work to be done" -
     which describes all of the mutations and test runs that need to be
@@ -96,7 +97,11 @@ def init(config_file, session_file):
             log.info(" - %s: %s", directory, ", ".join(sorted(files)))
 
     with use_db(session_file) as database:
-        cosmic_ray.commands.init(modules, database, operators_cfg)
+        if database.num_results>0:
+            sys.exit(ExitCode.OK)
+        elif database.num_results==0 or force:
+            cosmic_ray.commands.init(modules, database, operators_cfg)
+            
 
     sys.exit(ExitCode.OK)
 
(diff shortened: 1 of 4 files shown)


src/cosmic_ray/cli.py
```diff
@@ -96,7 +97,11 @@ def init(config_file, session_file):
             log.info(" - %s: %s", directory, ", ".join(sorted(files)))
 
     with use_db(session_file) as database:
-        cosmic_ray.commands.init(modules, database, operators_cfg)
+        if database.num_results>0:
```
abingham, line 100: I think this logic is incorrect. If `database.num_results` is greater than zero, it will always exit, event if `force` is `True`. What we want is for initialization to occur if `force` is `True`, no matter what the state of the existing database is.

muhd360:
Ok will try

tests/unittests/test_command_line_processing.py
```diff
@@ -59,11 +60,24 @@ def test_non_existent_session_file_returns_EX_NOINPUT(local_unittest_config):
     assert cosmic_ray.cli.main(["exec", str(local_unittest_config), "foo.session"]) == ExitCode.NO_INPUT
 
 
-def test_non_existent_config_file_returns_EX_NOINPUT(session, local_unittest_config):
-    cosmic_ray.cli.main(["init", local_unittest_config, str(session)])
+def test_non_existent_config_file_returns_EX_NOINPUT(session, local_unittest_config,force):
```
abingham, line 63: This test (and all tests which take a `force` parameter) don't work because you don't have a fixture providing the value for `force`. In all cases, I don't think the tests actually need to use a fixture for `force`. You can simply pass the correct value - `True` or `False` - into `cosmic_ray.cli.main`.

tests/unittests/test_command_line_processing.py
```diff
@@ -59,11 +60,24 @@ def test_non_existent_session_file_returns_EX_NOINPUT(local_unittest_config):
...
+    cosmic_ray.cli.main(["init", local_unittest_config, str(session),force])
     assert cosmic_ray.cli.main(["exec", "no-such-file", str(session)]) == ExitCode.CONFIG
 
 
+
+def test_init_with_existing_results_no_force(session, local_unittest_config,force):
+    """Test that init exits without reinitializing when results exist and force=False"""
+    with use_db(session) as database:
+        database.num_results = 1  # Simulate existing results
```
abingham, line 72: You can't just set `num_results` like this. The `WorkDB` class won't allow it. 

You clearly haven't run these tests because none of them run. Please don't resubmit until the tests actually run.

abingham (changes requested):

tests/unittests/test_command_line_processing.py
muhd360: sorry i commited the wrong branch will fix the changes

muhd360:
hi @abingham for the case when force is false what will the test case reflect is creating a mockdb the right kind of approach???

muhd360:
@abingham ???

abingham:
> hi @abingham for the case when force is false what will the test case reflect is creating a mockdb the right kind of approach???

A test for the `force == False` case would not require a mock database. You could create a normal workdb, get some results into it, and then verify that running `init` with `force == False` aborts and does not re-init the database, i.e. it would leave the database unchanged.

muhd360:
ok will do

muhd360:
hi @abingham both the tests are still failing can u pls check them once and guide me to the necessary steps so that i can fix my mistake i cant seem to figure it out

tests/unittests/test_command_line_processing.py
```diff
@@ -64,6 +66,56 @@ def test_non_existent_config_file_returns_EX_NOINPUT(session, local_unittest_con
...
+    # Verify initial database state has our test work items
+    initial_count = db.num_work_items
+    assert initial_count == len(work_items)
+
+
+
+    result = cosmic_ray.cli.main(["init", local_unittest_config, str(session)])
+    db=WorkDB(session,2)
+    final_count = db.num_results
+    assert final_count == initial_count  # Confirm work items are unchanged
```
abingham, line 101: You seem to be comparing the wrong things here. `final_count` is the number of *results*, i.e. the number of work items which have actually been executed. `initial_count` is the simply the total number of work items, whether or not they have results attached to them. You need to actually get some results into the workdb for this test to make any sense, and even then you need to make sure you compare the number of *results* before and after your init call.

muhd360:
thank u so much.....will do the necesarry changes.thx once again

muhd360:
hi i cant fix the behaviour of execute what can I do????

abingham:
This has gone a bit off the rails, and I don't have much free time to commit to sorting it out. I'll try to take a look at it this weekend, but I think it'll probably easiest if I just rewrite it.

abingham:
I've committed fixes for this issue. You can see the changes in 9e6eb65 and 408f417. Thanks so much for your work on this. As you can see, the change is essentially identical to what you implemented, but it was easier for me to write the tests myself than explain what to do.

PR #564, opened 2025-02-27
Add support for higher-order mutants

### LINKED ISSUE #296: Second- and higher-order mutants
There seems to be quite a lot of research which suggests that higher-order mutants (mutants containing more than one mutation) can be useful in several ways:

1) They can reduce the overall number of mutants to test.
2) They can reduce the likelihood of mutants being equivalent mutants

Some of the suggested mutation combination algorithms are very simple and reported to be effective.

This would seem to suggest that less, but more useful, work would be done by Cosmic Ray. 

Further reading:

- [Higher order mutation testing: A Systematic Literature Review](http://www.sciencedirect.com/science/article/pii/S1574013716301095)
- [An Empirical Evaluation of the First and Second Order Mutation Testing Strategies](http://ieeexplore.ieee.org/document/5463722/)
- [Overcoming the Equivalent Mutant Problem: A Systematic Literature Review and a Comparative Experiment of Second Order Mutation](http://ieeexplore.ieee.org/document/6613487/)

andre15silva:
This implementation adds support for higher-order mutants (multiple mutations applied simultaneously). Key changes include:

1. Added mutation-order config option to control mutation order
2. Refactored database schema to support multiple mutations per work item
3. Updated CLI to support testing multiple mutations at once
4. Added comprehensive tests for higher-order mutations
5. Added detailed documentation and examples

Closes #296

### DIFF
--- a/docs/source/concepts.rst
+++ b/docs/source/concepts.rst
@@ -187,3 +187,87 @@ timeout after 10 seconds, use:
    # config.toml
    [cosmic-ray]
    timeout = 10
+   
+Higher-order Mutants
+===================
+
+By default, Cosmic Ray applies a single mutation at a time, creating what are known as *first-order mutants*. 
+This means each test run tests exactly one mutation. However, Cosmic Ray also supports *higher-order mutants*, 
+which are produced by applying multiple mutations simultaneously.
+
+You can configure the order of mutations through the ``mutation-order`` configuration key. This specifies 
+the maximum number of mutations that will be applied in a single test run. For example, to specify that 
+tests should apply up to 2 mutations simultaneously, use:
+
+.. code-block:: ini
+
+   # config.toml
+   [cosmic-ray]
+   mutation-order = 2
+   
+When a value greater than 1 is specified, Cosmic Ray will generate work items for all possible combinations of 
+mutations up to the specified order. For example, with ``mutation-order = 2``, it will generate all single mutations 
+(first-order) and all pairs of mutations (second-order).
+
+Higher-order mutants can be useful for:
+
+1. Testing the robustness of your test suite against more complex mutations
+2. Finding interactions between mutations that might not be detected with first-order mutants
+3. Reducing test execution time by testing multiple mutations at once
+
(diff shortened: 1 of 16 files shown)


andre15silva:
FYI, this is WIP but in case you have something to add feel free too!

Disclaimer, most of the code is written by Claude Code

abingham:
I've only been able to look at this briefly, but it looks like a really good piece of work. In particular, I think the work on the `WorkDB` and the core ability to have multiple mutations per job is a critical addition to CR for handling higher order mutations.

My main concern is that this builds a particular approach to higher-order mutation into the core of CR, and I'm reluctant to make the core any more complex than it needs to be. While I'm happy to update the core to support multiple mutations for a job, I would rather leave any particular strategy for generating higher-order mutations to filter programs. With the approach I envision, the `init` command would only populate the `WorkDB` with order-1 mutations. Then you would run a filter to add higher-order mutations to the `WorkDB` (perhaps disabling order-1 mutations in the process), using whatever strategy you want for selecting them. I don't think it's big conceptual shift from what you've created, it just moves the responsibility around a bit and doesn't "bless" any particular HOM strategy.

Does this make sense? Does it fail to take anything into account? I'm quite (perhaps too) conservative with core changes to CR, so let me know if doing HOM with filters would be too difficult or awkward for some reason. I think we'll be able to use most of your implementation in any event.

And thanks for doing this! I don't have much time to put into CR right now, and it's great to see some activity in the project.

andre15silva:
@abingham thanks for the comment!

I think it's a good idea to split into different stages, I'll try to make that change after getting the core idea to work.

Quick question for you: I am getting some NO_TEST worker outcomes in some high-order mutations. I have only been able to replicate this on my own project (with custom mutation operators), but haven't been able to understand why these occur.

AFAIU, this outcome should only happen if there is no test to be run (which should not be the case). Any idea why this might happen?

abingham:
I'll try to look at these things this weekend. I'm extremely busy at work, but know that I'll get to it when I can.

Issue #565, opened 2025-08-17
Pipe operator mutation in the type hint of annotated assignment survives


ashres:
```python
def example_function(n: int | str) -> str | int:
    value: str | int = (
        "Hello"
        if isinstance(n, int) and (n | 0xFF) == 0x1FF
        else 1
        if isinstance(n, str) and n == "hi"
        else 2
    )

    return value
```

In the contrived example above, the pipe operator is found in 4 different locations:

1. In the function's argument's type hint
2. In the function's return type's type hint
3. In the local variable `value` annotated assignment's type hint
4. In the first conditional as a bitwise `OR` 

Because it sees the pipe as a binary operator, CR will mutate each of these instances.  

Mutants for \# 4  would be killed because mutations would actually impact runtime behavior.  So, no need to focus on that case.  The others, however, are type hints and shouldn't have runtime implications. 

Replacing the pipe operator with another binary operator to \# 1 and \# 2 cause `TypeError`s to be raised (when tested with `python 3.13` and `python 3.10`):

```
  File "/Users/me/Documents/code/python/test-bed/main.py", line 21, in <module>
    def example_function(n: int & str) -> str | int:
                      ~~~~^~~~~
TypeError: unsupported operand type(s) for &: 'type' and 'type'
```

```
  File "/Users/me/Documents/code/python/test-bed/main.py", line 21, in <module>
    def example_function(n: int | str) -> str & int:
                                    ~~~~^~~~~
TypeError: unsupported operand type(s) for &: 'type' and 'type'
```

Interestingly, doing the same for \# 3 causes the unit test to run successfully:

```
value: str & int = ...
```

I suspect we get this behavior because [local variable annotations are not stored at runtime](https://docs.python.org/3/glossary.html#term-annotation), while the others are.

So, CR should avoid mutating the pipe operator within the type hints of annotated assignments to avoid false-positive mutant survivals.

abingham:
@ashres Is this resolved now?

PR #566, opened 2025-08-17
Issue #565 - Avoid mutating pipe operator in assignment annotations

ashres:
Currently, mutating pipes in the annotations of assignments create false mutant survivals because there is not runtime implications (errors or otherwise).

This patch will prevent the pipe character within the annotations of assignments from being seen as a binary operator, thus preventing mutations of that character and eliminating the false-positive mutant survival.


abingham:
I'm traveling right now, but I'll try to look at this as soon as I get back.

abingham:
Thanks!

PR #591, opened 2026-04-02
fix(filters): remove noisy pragma-no-mutate stdout spam

abingham:
Refs: #586


mattcann12:
@abingham Hey any update on this working itself into master?

PR #592, opened 2026-04-02
replace qprompt with click for new-config

abingham:


src/cosmic_ray/commands/new_config.py
```diff
@@ -2,7 +2,7 @@
 
 import os.path
 
-import qprompt
+import click
```
liquidpele, line 5: recommend moving the click import into the function that uses it, unless you want it as a hard requirement, if you do then don't forget to add it to the toml where you removed qprompt.

abingham: click has been a hard requirement of cosmic ray for years, so it's already in the pyproject.toml dependencies. Honestly, I've never been a big fan of this feature of `new_config()`, so this is making me rethinking if the prompting should a) exist and b) be put in the CLI more directly.

PR #595, opened 2026-04-16
Update exec command to allow setting log level

liquidpele:
This allows using --verbosity=INFO with exec so that you can watch the logs as the system is creating mutations.


kvandeh:
Finally

abingham (approved):

PR #603, opened 2026-08-02
Kill the whole test process group on timeout

### LINKED ISSUE #600: run_tests() timeout doesn't reliably kill hung test subprocesses — orphaned child can run unbounded and exhaust system memory
Version: cosmic-ray 8.4.6, macOS (Tahoe 26.5.2), Python 3.11, test-command = uv run pytest ...

What happened: During a ~1,500-job local mutation run, at least one mutation made the test-command hang instead of finishing or failing. run_tests() in cosmic_ray/testing.py relies on subprocess.run(shlex.split(command), timeout=timeout, ...), which per Python's own documented behavior only kills the immediate child process on timeout — not any further descendants that child may have spawned. In our case the immediate child (uv) appears to have left behind an orphaned pytest process (confirmed via ps: reparented to PID 1/launchd, meaning its original parent was gone but it kept running). That orphan then grew to ~25GB RSS over several minutes with nothing supervising or bounding it, eventually exhausting available memory. This happened twice on two separate long runs, and both times macOS's own watchdog force-restarted the machine (watchdog timeout: no checkins from watchdogd / userspace watchdog timeout: no successful checkins from WindowServer) because the system had become unresponsive under memory pressure — i.e., a hung mutation subprocess was able to crash the host OS, not just fail the job.

Evidence the timeout mechanism is only partially effective: in the same session's database, 26 jobs did correctly record output == "timeout" (TestOutcome.KILLED), so the happy path works. The failure is specific to whatever this one hanging mutation does — consistent with a child-of-child process surviving Popen.kill().

Not yet identified: the exact mutation/line responsible. Two plausible candidates in our own code (two while-loop dedup patterns) were tested directly under a hard-bounded harness and ruled out, so it's something else in a ~1,500-mutation surface.

Suggested direction: spawn the test command in its own process group (preexec_fn=os.setsid on POSIX) and kill the whole group (os.killpg) on timeout, rather than killing only the tracked PID — the standard fix for this class of Python subprocess timeout gap. Happy to share the workaround wrapper we built (puts the test command in its own process group, hard-kills on a wall-time or RSS budget) if useful as a reference.

Sanjays2402:
Closes #600

`subprocess.run()` only kills the immediate child on timeout, so a test command that spawns further processes (e.g. `uv run pytest`) leaves orphaned descendants running unsupervised — in the reported case one grew to ~25GB RSS and took the host down. `run_tests()` now uses `start_new_session=True` and kills the whole process group on timeout, as suggested in the issue thread. Outcome classification is unchanged.

New `tests/unittests/test_testing.py` spawns a grandchild from a command that times out and asserts it does not survive; it fails without the fix and passes with it.

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.


abingham (changes requested):
This looks good. Please rebase it on main to get rid of the error in the test suite (which I accidentally introduced). Once everything is passing I'll merge it in.

